PR #193, opened 2026-04-10
Add DataFrame-based ingestion support to BERIL AI agent ingest skill

khanahmedm:
This PR enhances the BERIL AI agent ingest skill by documenting support for DataFrame-based ingestion using the `dataframes` parameter in the `ingest()` API.

Previously, the skill only described file-based ingestion workflows that rely on:
- local files
- MinIO uploads
- pandas-based chunking for large datasets

This update introduces an alternative ingestion mode that allows users to pass in-memory DataFrames directly to the ingestion framework.

### DIFF
--- a/.claude/skills/berdl-ingest/SKILL.md
+++ b/.claude/skills/berdl-ingest/SKILL.md
@@ -490,14 +490,14 @@ from data_lakehouse_ingest import ingest
 
 ingest(
     config,
-    df_override={
+    dataframes={
         "table_name": dataframe
     }
 )
 ```
 
 **Key Behavior**
-* `df_override` maps **table name → DataFrame**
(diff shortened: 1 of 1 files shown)


briehl:
@khanahmedm @mikacashman Is this still useful for you all?

khanahmedm:
> @briehl @mikacashman Is this still useful for you all?

DataFrame-based ingestion is one of the features available in the ingest library. It enables users to load data into the data lakehouse even when they do not have a load-ready file, which was previously required by the library. For example, if a user has transformed data stored in a DataFrame but does not want to create an intermediate load-ready file, they can use this feature to directly load the data into the lakehouse.

Since BERIL uses the ingest library, that's why I added this PR.

I hope this helps.

PR #215, opened 2026-04-28
chat: ChatSession and ChatMessage tables (2/7)

briehl:
Second in a 7-PR series adding an LLM chat interface. Adds the
persistent storage that later PRs use for transcripts and resume
tokens. Stacks on top of #214 (PR #1).

## What this PR adds

- **`ChatSession`**: `owner_id` (FK to `beril_user` with cascade
  delete), `sdk_session_id` (resume handle from the Agent SDK),
  `provider_id` (free string matched against the registry at runtime —
  adding a new provider is config + code, not a migration), `model`,
  `title`, optional `project_id`, `archived` flag, `created_at` /
  `last_active_at`.
- **`ChatMessage`**: `session_id` (cascade delete), role enum
  (`user`/`assistant`/`tool`/`tool_result`), JSONB `content` (text +
  optional `tool_activity` + optional `error`), composite index on
  `(session_id, created_at)` for transcript ordering.
- `JSONCol = JSON().with_variant(JSONB(), "postgresql")` so
  SQLite-backed unit tests work too.
- `chat_sessions` back-relations on `BerilUser` (cascade delete) and
  `UserProject` (SET NULL on project deletion — chats outlive projects).
- Alembic revision `c3d4e5f6a1b2` matches the existing idempotent style.

## Tests

11 new model tests covering creation, relationships, cascade behavior,
free-form `provider_id`, and chronological ordering. Full suite: 572
passing.

## Stack context

This is **2 of 7**. PR #1 (#214) is merged. Subsequent branches
(`chat/3` … `chat/7`) sit on top locally and will be rebased onto each
predecessor's merged form.

PR #218, opened 2026-04-29
chat: provider abstraction + Anthropic-compatible implementation (3/7)

briehl:
Third in a 7-PR series adding an LLM chat interface. Defines the
provider Protocol all later PRs program against, and the one
implementation needed for the shipped catalog (Anthropic + CBORG).
Stacks on top of #215.

## What this PR adds

- **`ChatProvider` Protocol** with a single async `run_turn()` that
  yields a `TurnEvent` stream. Event types: `SessionInitialized`,
  `TextDelta`, `ToolCall`, `ToolResult`, `TurnComplete`, `ErrorEvent`.
  The stream shape is the contract between provider implementations
  and the service layer that lands in PR #4.
- **`Credentials`** dataclass with `require()` — declarative match
  against the provider config's `credential_fields`. Bedrock-style
  multi-field credentials slot in without API changes.
- **`AnthropicCompatibleProvider`**: wraps `claude_agent_sdk.query()`
  and translates `SystemMessage`/`AssistantMessage`/`ResultMessage` and
  blocks into `TurnEvent`s. Per-turn `env`
  (`ANTHROPIC_BASE_URL` plus either `ANTHROPIC_API_KEY` or
  `ANTHROPIC_AUTH_TOKEN` depending on the provider's `auth_style`)
  keeps user keys out of the parent process env. The unused auth env
  var is blanked so a stale server-level default cannot shadow the
  user's token.
- **Lazy registry** built from `get_chat_config()`;
  `get_provider("anthropic")` / `get_provider("cborg")` return cached
  instances.

## Tests

21 tests covering frame translation (each block type), missing
credentials, the `auth_style` branch, registry lookup/caching/reset,
and a synthesized `TurnComplete` when the SDK ends without one. Full
suite: 607 passing.

Note: the SDK is not exercised end-to-end here; the service layer that
calls it (and a real subprocess) lands in PR #4.

## Stack context

This is **3 of 7**. PRs #1 (#214) and #2 (#215) are merged.

### DIFF
--- a/ui/app/chat/providers/__init__.py
+++ b/ui/app/chat/providers/__init__.py
@@ -1,0 +1,57 @@
+"""Chat provider registry.
+
+The registry maps ``provider_id`` strings (as stored on
+``ChatSession.provider_id``) to concrete provider instances. Membership is
+driven by the chat config YAML: each entry in ``providers`` becomes one
+registry entry, selected by its ``wire_protocol``.
+
+Adding a new wire protocol (e.g. OpenAI, Bedrock) is a code change — add the
+implementation and extend :func:`_build_provider` — but adding a new provider
+that speaks an existing wire protocol is a config edit only.
+"""
+
+from __future__ import annotations
+
+from app.chat.config import ProviderConfig, get_chat_config
+from app.chat.providers.anthropic_compatible import AnthropicCompatibleProvider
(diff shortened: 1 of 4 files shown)


briehl:
Fixed various review issues, merging.

PR #226, opened 2026-04-30
Chat: SSE streaming endpoint

briehl:
Fifth in the chat-feature PR series. Adds POST /api/chat/{id}/stream returning Server-Sent Events so the frontend can render assistant responses as they're generated rather than waiting for the full reply.

- ui/app/chat/sse.py: event_to_frame() maps each TurnEvent to the {event, data} dict EventSourceResponse expects. turn_complete_frame() emits a final ``turn_persisted`` frame carrying the newly-committed assistant message id, so the client can reconcile the in-flight optimistic message with the server-canonical one.
- service.stream_turn(): async generator that runs the turn, persists the user message before dispatch, yields SSE frames as provider events arrive, and persists the assistant message exactly once at the end. If the client disconnects mid-stream, the finally branch still persists whatever was accumulated (marked error="stream interrupted") so a reload doesn't lose partial work.
- routes.chat_stream(): pre-flight (auth, ownership, credentials, cap) runs synchronously and returns plain JSON errors (401/403/404/400/429) before the SSE response starts - so the React client can handle failures with the same machinery as the non-streaming endpoint rather than parsing in-band error frames. The concurrency slot is reserved before the response is returned and released in the generator's finally.
- 21 new tests: 8 SSE frame mapping unit tests, 13 streaming route e2e tests (auth, validation, cap, happy-path frame sequence, persistence, sdk_session_id capture, provider error paths, slot release after stream end).

Note: this PR makes streaming work over the wire but there's still no UI consuming it - the frontend will be in PR #7 after the session-management pages in PR #6.

### DIFF
--- a/ui/app/chat/service.py
+++ b/ui/app/chat/service.py
@@ -152,8 +152,14 @@ async def stream_turn(
         # failure — to avoid duplicating the assistant row.
         if not persist_started:
             result.assistant_text = "".join(text_parts)
-            if not stream_completed and result.error is None:
-                result.error = "stream interrupted"
+            if not stream_completed:
+                # Don't promote a partial provider session to the resume
+                # handle — the next turn must rebuild from the DB transcript
+                # rather than resume from a conversation we only partially
+                # captured locally.
+                result.sdk_session_id = None
+                if result.error is None:
+                    result.error = "stream interrupted"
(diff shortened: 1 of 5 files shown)


briehl:
Reviews and fixes done offline, merging.

PR #238, opened 2026-05-01
chat: React 19 + TypeScript chat component (7/7)

briehl:
## Summary

Final PR in the chat-feature series. Adds the user-facing chat UI as a
self-contained Vite-built ES module that the FastAPI session page mounts
as an island.

- `ui/components/chat/` — standalone TypeScript sub-package (React 19 +
  Vite). Output is a single ES module written to `ui/app/static/chat/chat.js`
  (the output dir is gitignored — built locally with `npm run build`).
- `types.ts` is the cross-boundary contract with `app/chat/sse.py`. If
  one side renames a field, the other breaks at runtime.
- `sseClient.ts` does POST + fetch + ReadableStream consumption. Native
  `EventSource` can't POST (we send the API key in the body), so we
  frame SSE by hand. Normalizes CRLF to LF before splitting because
  `sse-starlette` emits `\r\n` separators.
- `credentialsStore.ts` keeps API keys in `localStorage` only — they
  never leave the browser between turns.
- `App.tsx` uses `flushSync` on row insertion so `text_delta` updaters
  always find the assistant row in `prev`. `inFlightRef` guards against
  double-submit. Frame folding is by message id (not "any pending
  message"), so resumes work.
- `vite.config.ts` includes `define: { 'process.env.NODE_ENV': '"production"' }`
  — without that the React production-vs-dev branch reference survives
  into the bundle and throws `ReferenceError: process is not defined`.

After merge, build with:

```
cd ui/components/chat && npm install && npm run build
```

Then `/chat/{id}` mounts the React component, sends user messages to the
SSE endpoint from #226, and streams the response. The SSR transcript
stays in the DOM as the no-JS fallback and is hidden once the React
island is alive.

Stacked on top of #237 (merged).

## Test plan
- [ ] `cd ui/components/chat && npm install && npm run build` produces
      `ui/app/static/chat/chat.js`.
- [ ] `/chat/{id}` loads, accepts a message, streams the response.
- [ ] API key persists across reloads via localStorage.
- [ ] No-JS fallback (SSR transcript) renders if the bundle is missing.
- [ ] `cd ui && uv run pytest tests/ -q --no-cov` — 712 tests still pass
      (no Python changes in this PR).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -87,6 +87,14 @@ ui/data/cache.json
 ui/data/indexdir/
 ui/user_projects/
 
+# Chat component build output (generated from ui/components/chat/)
+ui/app/static/chat/
+
+# Frontend build artifacts & deps
+node_modules/
+ui/components/*/dist/
+.vite/
+
 # Python virtual environments
 .venv/
 .venv-berdl/
--- a/ui/components/chat/.gitignore
+++ b/ui/components/chat/.gitignore
@@ -1,0 +1,4 @@
+node_modules/
+dist/
+.vite/
+*.log
--- a/ui/components/chat/README.md
+++ b/ui/components/chat/README.md
@@ -1,0 +1,104 @@
+# BERIL Chat Component
+
+Self-contained React/TypeScript chat UI for the BERIL Research Observatory.
+Built with Vite, output as a single ES module to `ui/app/static/chat/chat.js`,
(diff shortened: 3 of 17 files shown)


briehl:
TODO: add build step to the GHA image build process.

PR #239, opened 2026-05-01
feat: OpenViking knowledge context layer

dileep-kishore:
## Summary

- Adds a curated OpenViking-backed knowledge context layer at `observatory_context/` plus thin CLIs under `knowledge/scripts/` (`ingest_context.py`, `knowledge_query.py`, `smoke_ingest_openviking.py`) for ingesting and querying BERIL project/docs context.
- Stable per-project URIs (`viking://resources/projects/<id>/`), a separate `project_index/` URI to avoid target-lock overlap, and a sha256 manifest under `knowledge/state/` driving incremental `--changed` ingestion with symmetric stale-target removal across `--all`/`--project`/`--docs`.
- Local OpenViking workflow via `knowledge/openviking/ov.conf.example` (OpenRouter embedding + Gemini Flash VLM), repo-local workspace, plus a `knowledge-context` Claude skill and integration notes in `pitfall-capture`, `suggest-research`, and `synthesize`.

### Ingest UX & reliability

- Rich CLI progress for `ingest_context.py` and a fast OpenViking health probe before opening the client, so misconfigured servers fail loudly up front instead of mid-ingest.
- `wait_processed` now polls `get_status` per queue with a 1h safety deadline, replacing the SDK single blocking POST that was hitting httpx 600s read timeouts on slow VLM backends. Three per-queue progress bars (Embedding / Semantic / Semantic-Nodes) replace the old single status line.
- `--limit N` on `--all` / `--changed` caps how many projects are ingested per run (handy for rate-limited VLM providers like CBORG at 20 RPM). Partial runs merge into the existing manifest so a follow-up `--changed` picks up the skipped projects; stale-target reconciliation is suppressed under `--limit` so a partial run never persists half-finished state.
- `ov.conf.example` aligned with the current config (OpenRouter dense embedding, Gemini Flash VLM) and quieted default logs.
- Test suite trimmed to focused cases covering the golden ingest paths, the regressions that actually broke (4/2 progress counting, sequential per-project waits, queue-drain detection), and the rendering primitives.

## Test plan

- [ ] `uv run --group knowledge pytest knowledge/tests -q`
- [ ] `uv run --group knowledge python knowledge/scripts/ingest_context.py --help`
- [ ] `uv run --group knowledge python knowledge/scripts/knowledge_query.py --help`
- [ ] `uv run --group knowledge python knowledge/scripts/ingest_context.py --changed --limit 3` against a running local OpenViking
- [ ] Optional smoke (requires running local OpenViking): `uv run --group knowledge python knowledge/scripts/smoke_ingest_openviking.py`


briehl:
This all seems to work locally, but I'm not sure if having the skills in will cause problems without having a deployment in place. I'm building a Dockerfile to get it installed on SPIN, we can merge once that's up and stable.

dileep-kishore:
It could still fall back to using the project files if OpenViking doesn't exist

cmungall:
my understanding is that we won't be deploying this tomorrow

briehl:
@cmungall Right, not deployed for today. Sorry for the late reply.

dileep-kishore:
Merged latest `main` (192 commits: Iceberg/Polaris migration, per-project memory overhaul, `wiki/`→`atlas/`). Only 3 trivial conflicts (`pyproject.toml`, `.gitignore`, `suggest-research/SKILL.md`), all resolved keep-both; lockfile relocked.

Also aligned the context layer with main's memory overhaul: ingest now covers `projects/<id>/memories/*.md`, and the pitfall-capture note points at per-project memories instead of the now-frozen `docs/pitfalls.md`.

🚧 Testing in progress to confirm functionality is preserved post-merge (`knowledge/tests` green: 37 passed; verifying end-to-end ingest next).

dileep-kishore:
Tested the context layer against a live OpenViking server: ingest of 2 new projects, all query primitives, the 4 skills that read context (berdl_start, pitfall-capture, suggest-research, synthesize), and degraded fallback. No lakehouse needed.

Found and fixed two bugs:

- fix(query): bad input (empty query, missing URI, bad --filter/--since) printed a full traceback. Now prints a clean error and exits 1.
- fix(ingest): `--project ""` reported a false "Ingest finished" success. Now rejected before touching the server.

Added regression tests for both.

mikacashman (approved):
LGTM - should test additional skills to see if there are any unexpected interactions.  To do as we go

PR #242, opened 2026-05-01
fix: use live BERDL discovery helpers

### LINKED ISSUE #229: Use BERDL specific function for spark queries
Reference: https://github.com/BERDataLakehouse/spark_notebook/blob/main/notebook_utils/berdl_notebook_utils/__init__.py

### LINKED ISSUE #240: Remove collections.md and related files in favor of live Spark queries for tenant/table access
The collections.md is currently used as a reference for tenants/tables, but it doesn't accurately reflect what a given user actually has access to. This creates confusion when users (or the agent) treat it as the source of truth.

### LINKED ISSUE #241: /berdl skill hardcoded to REST API — environment detection is overridden by skill body
berdl_start runs detect_berdl_environment.py to determine whether the agent is on the BERDL hub. Detection works correctly. The problem is downstream: queries still go through --berdl-proxy (or REST) even when the agent is on-cluster and should be using Spark directly.

dileep-kishore:
## Summary

- switch BERDL collection discovery to access-aware `berdl_notebook_utils` helpers
- remove `docs/collections.md` as a static access source of truth while keeping the UI snapshot tracked as a display/fallback artifact
- update BERDL skills and docs to use Spark SQL instead of REST/MCP query operations
- route BERDL query guidance by environment: direct `spark.sql(query)` on BERDL JupyterHub, `--berdl-proxy` only for off-cluster local execution

## Testing

- `cd ui && uv run --python 3.11 pytest tests/test_discover_berdl_collections.py tests/test_dataloader.py::TestParseCollections -q`
- `cd ui && uv run --python 3.11 --with greenlet pytest tests/test_main.py::TestCollectionsRoute tests/test_wiki_lint.py -q`
- `cd ui && uv run --python 3.11 ruff check app tests`
- `python3 -m py_compile scripts/discover_berdl_collections.py && git diff --check`

Fixes #229
Fixes #240
Fixes #241


mikacashman (commented):
(test comment - disregard)

mikacashman (commented):
(test comment - disregard)

mikacashman (commented):
(test comment - disregard)

mikacashman (commented):
(test comment - disregard)

mikacashman (commented):
The on-cluster path using `berdl_notebook_utils` is correct and should remain the primary path. However, `discover_berdl_collections.py` is now unusable off-cluster because `berdl_notebook_utils` requires JupyterHub environment variables (`BERDL_POD_IP`, `SPARK_MASTER_URL`, `BERDL_HIVE_METASTORE_URI`, etc.) that are not set in the local `.env`. Before this PR the script had a working REST API path that only needed `KBASE_AUTH_TOKEN`.

The suggestions below restore a REST fallback that activates **only** when `berdl_notebook_utils` cannot be imported (i.e. off-cluster). On-cluster, `berdl_notebook_utils` is always tried first and the REST path is never reached.

Six suggestions total (two for `main()` because it spans two diff hunks):
1. Restore `import os`, `urllib.error/request`, and `DEFAULT_BASE_URL`
2. Restore `read_auth_token()` and `_post_json()` helper functions
3. Fix `_load_berdl_helpers()` to return `None` off-cluster instead of raising
4. Add REST fallback branch to `discover_collections()`
5a. Add `--base-url` and `--timeout` args to `main()`
5b. Add `--env-file`, wire `token` through, fix error handling in `main()`

Edit: Will address changes in a new PR

PR #246, opened 2026-05-04
fix: restore off-cluster REST fallback in discover_berdl_collections

mikacashman:
berdl_notebook_utils requires full JupyterHub env vars that aren't set
  locally, breaking the script off-cluster. Restore the REST API path as
  an automatic fallback when berdl_notebook_utils cannot be imported.

  On-cluster: berdl_notebook_utils is always used (access-aware, unchanged).
  Off-cluster: falls back to REST API with KBASE_AUTH_TOKEN from .env.

  - _load_berdl_helpers() returns None instead of raising on ImportError
  - discover_collections() dispatches to helpers or REST based on that
  - Restore read_auth_token(), _post_json(), DEFAULT_BASE_URL
  - Restore --base-url, --timeout, --env-file CLI args to main()

### DIFF
--- a/scripts/discover_berdl_collections.py
+++ b/scripts/discover_berdl_collections.py
@@ -5,12 +5,17 @@ from __future__ import annotations
 
 import argparse
 import json
+import os
 import sys
 import tempfile
+import urllib.error
+import urllib.request
 from datetime import datetime, timezone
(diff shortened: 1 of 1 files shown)


dileep-kishore (approved):
Looks good to me!

PR #247, opened 2026-05-05
Skill/local ingest

mikacashman:
Moves skill `berdl-ingest` to `berdl-ingest-remote` and creates new `berdl-ingest` skill to perform ingest on-cluster (in K-BERDL JupyterHub environment).


briehl (approved):

briehl:
Failing tests are due to another reason. I'll put in a fix later.

PR #255, opened 2026-05-06
Harden BERDL skills: env check, live discovery, single source of truth

dileep-kishore:
## Summary

The BERDL skill suite had three compounding failure modes:

1. **Inconsistent environment detection** — only `/berdl_start` ran detection. Other skills hallucinated script names like `berdl_detect` and silently fell back to remote workflows when on-cluster.
2. **Hardcoded inventory drift** — skill markdown and `docs/schemas/` (~3,200 lines) duplicated structural facts that drift from the live database. Agents trusted the docs over reality.
3. **Helpers misused** — `berdl_notebook_utils.get_databases()` defaults to `return_json=True` (returns a JSON *string*). Every skill called it without `return_json=False` and treated the return as a list. `"foo" in '["foobar"]'` returned True; `display()` printed garbage.

This PR fixes all three.

### What's in here

- **`scripts/berdl_env.py`** — single canonical Step 0 for every BERDL skill. Auto-starts pproxy when SSH tunnels are up but pproxy isn't; prints exact `ssh -f -N -D` commands when tunnels are missing. CLI (`--check`, `--json`) and Python API (`require_environment`, `BerdlEnvironmentError`). 8 pytest unit tests.
- **Every BERDL skill** (`berdl_start`, `berdl`, `berdl-query`, `berdl-minio`, `berdl-discover`, `berdl-ingest`, `berdl-ingest-remote`) now calls `berdl_env.py --check` as Step 0. The location it reports (`on-cluster` / `off-cluster`) selects the execution path — skills no longer infer it from import errors or guesses.
- **All `get_databases` / `get_tables` / `get_table_schema` calls fixed** to pass `return_json=False`. The silent `namespace in get_databases()` substring bug in `berdl-ingest` is gone.
- **`berdl-discover` rewritten** as a thin wrapper around `get_table_schema(detailed=True)` + `spark.sql("DESCRIBE EXTENDED ...")`. Drops module-file generation; proposes pitfall additions to `docs/pitfalls.md` instead.
- **`docs/schemas/` deleted** (18 files, ~3,200 lines). Structural info comes from the live database; non-derivable gotchas migrated to per-database H2 sections in `docs/pitfalls.md`.
- **Off-cluster mechanics separated.** Proxy/tunnel/`mc`/`.venv-berdl` content extracted out of `docs/pitfalls.md` and `docs/performance.md` into `.claude/skills/berdl-query/references/off-cluster-mechanics.md` and `.claude/skills/berdl-ingest/references/on-cluster-bypass.md`. General docs become environment-agnostic.
- **`berdl/modules/{pangenome,biochemistry,paperblast}.md` deleted** (duplicated `docs/schemas/` and the live database). `query-patterns.md` and `cross-database.md` kept, row counts stripped, example callout added.
- **Hardcoded scale claims removed** from `docs/overview.md`, `PROJECT.md`, `docs/research_ideas.md`. Analysis-output counts kept; database/genome/gene inventory counts stripped (they go stale).
- **`tools/verify_skill_hardening.sh`** — 7-check static suite enforcing the invariants. Run anytime to re-verify.

## Test plan

- [x] `python -m pytest tests/scripts/test_berdl_env.py -v` — 8/8 passing
- [x] `bash tools/verify_skill_hardening.sh` — all 7 checks passing
- [x] On-cluster smoke (manual): `python scripts/berdl_env.py --check` should report `Location: on-cluster`, exit 0
- [x] Off-cluster smoke (manual, full proxy chain): same, exit 0
- [ ] Off-cluster smoke (manual, missing tunnels): exit 1 with exact SSH commands printed
- [ ] `/berdl_start` end-to-end: Phase 1.5 passes, Phase 1.6 prints live tenant-grouped DB list (no hardcoded numbers)
- [ ] `/berdl-discover <db>`: produces schema via `get_table_schema(detailed=True)` + `DESCRIBE EXTENDED`; does not write a module file


mikacashman:
Off-cluster `/berdl_start` wasn't able to look up the tenant information blocking on asking for the user to open a notebook and start spark (the old method we used).  May want the environmental check script to invoke the get_spark_session.py script which auto-spawns the JH server via berdl-remote.

PR #274, opened 2026-05-14
update URL in readme to the live site

briehl:
The previous URLs - http://beril-observatory.knowledge-engine.development.svc.spin.nersc.org/ should now be https://beril.kbase.us 

Hooray for SSL certs and brief URLs!


briehl:
This one's not exactly controversial, and already causing confusion, so I'll just merge.

PR #277, opened 2026-05-21
Add KBase auth provider

briehl:
First step toward handling KBase Lakehouse data within the UI (and, later, for handling user-owned data in KBERDL) is adding an auth layer.

This does that by adding a pluggable infrastructure that can be extended to provide other auth layers if other projects (phenix, OPAL, etc) need them. Probably, they won't, but the option is now there.

Right now, this doesn't do much except allow looking up a KBase auth token on request. Only certain requests will need it, so it's not always necessary - the `/health` endpoint is the only one in this PR with the extension.

Next step will be adding some flavor of a service token and config so the app can talk to KBERDL.


briehl:
This looks a lot bigger than it is, most changes are in `uv.lock`.

I should also mention here that this bumps the app to use Python 3.12, as that's what the KBase auth client needs.

dileep-kishore (approved):

PR #286, opened 2026-05-29
fix(ingest): complete Iceberg/Polaris migration in ingest code paths

mikacashman:
Finishes the Delta->Iceberg migration in the actual ingest code (prior commits migrated only the prose). Changes across both ingest skills and ingest_lib.py:

- Drop the explicit silver_base everywhere; Polaris owns the physical location. Tables are now queried as {tenant}.{dataset}.{table}.
- Use dotted Iceberg namespaces: {tenant}.{dataset} for shared tenants, my.{dataset} for the personal catalog. User-tenant configs now pass the bare dataset (the pipeline prepends "my" itself; prefixing here gave my.my.{dataset}).
- Register both get_minio_client and get_s3_client on the stub, since the Iceberg build of data_lakehouse_ingest looks up the renamed get_s3_client.
- Rewrite the in-cluster Setup to build Spark Connect + MinIO directly from env vars instead of ingest_lib's off-cluster proxy helpers (on-cluster bypass pattern).
- Fix verify_ingest to quote each namespace segment separately so Spark doesn't look for a schema literally named "tenant.dataset".
- Replace Delta/quarantine wording with Iceberg + progress-log guidance.

Co-authored by Claude Opus 4.8.


mikacashman:
Need to fix the on-cluster skill - please hold

dileep-kishore (changes requested):
Possible issue:

**SQLite path can silently drop the header row.** `export_sqlite` always writes a header, but `HAS_HEADER` is set from TSV/CSV-only detection and isn't pinned for SQLite. If it's left `False`, the column-name row is ingested as data and verify still passes (`expected = total_lines - 0` matches the over-counted rows). Suggest pinning it right after export: `HAS_HEADER = True  # export_sqlite always writes a header row`.

mikacashman:
Issue fixed as requested and addressed a few other bugs in the recent commit found in an AI review.  Ready for review @dileep-kishore

dileep-kishore (approved):
Looks good!

PR #298, opened 2026-06-17
add OpenViking config to app

briehl:
Start this OpenViking adapter from the app by simply adding configuration lines.


dileep-kishore (approved):
I like the OV embedding migration you set up!

PR #300, opened 2026-06-18
OpenViking deployment tooling

briehl:
Adds files related to OpenViking deployment on NERSC SPIN.


briehl:
This doesn't directly modify any BERIL behavior, but is helpful (necessary for me right now) for local development, so I'm merging it.

PR #301, opened 2026-06-24
Openviking adapter 2

briehl:
## Summary

Lets an ORCiD-authenticated BERIL user provision an account for themselves in a
deployed OpenViking (OV) instance and retrieve the resulting credentials, all
through the BERIL webapp. The OV `user_key` is minted by OV, stored encrypted at
rest in BERIL's Postgres, and only the owning user can fetch its plaintext.

### Adapter (the core feature — `d0ab794`)
- **Config**: new `ov_credential_key` (Fernet key) alongside the existing OV
  settings; `cryptography` is now an explicit dependency.
  — `ui/app/config.py`, `ui/pyproject.toml`, `.env.example`
- **Encryption at rest**: `encrypt_secret` / `decrypt_secret` (Fernet), with a
  clear `CredentialEncryptionError` when the key is unset/invalid.
  — `ui/app/crypto.py`
- **OV HTTP client**: async httpx client that authenticates as the ADMIN of the
  configured OV account (`Authorization: Bearer <ov_admin_key>`), scoped to
  `ov_account_id`. Parses OV's `{status, result, error}` envelope and raises a
  typed `OpenVikingError` carrying the HTTP status + OV error code (never the
  admin key). — `ui/app/clients/openviking.py`
- **Database**: one-to-one `OvUserCredential` ⇄ `BerilUser`, with
  `get_ov_credential` / `upsert_ov_credential` CRUD, and an Alembic migration
  chained off the current head. The plaintext key is never persisted.
  — `ui/app/db/models.py`, `ui/app/db/crud.py`,
  `ui/alembic/versions/20260618_d4e5f6a1b2c3_add_ov_user_credential.py`
- **Routes** (all gated by `require_user_api`; OV `user_id` is always the
  authenticated user's ORCiD, derived server-side):
  - `GET  /api/ov/health` — reachability probe to OV's public `/health`
  - `POST /api/ov/user` — idempotent create. Returns `created:false` if BERIL
    already holds a credential (no OV call). If OV reports the user already
    exists but BERIL has no stored key, returns **409** with an actionable
    message rather than silently regenerating.
  - `POST /api/ov/user/regenerate` — explicit, opt-in (destructive) key
    mint/rotation; the recovery path for the 409 case.
  - `GET  /api/ov/credentials` — returns the decrypted `user_key` to the owner.
  — `ui/app/routes/openviking.py`

### Local stack + deployment
- **docker-compose / nginx**: run BERIL UI + OpenViking + Postgres locally; an
  nginx proxy on `:8000` path-routes `/ov → openviking` and everything else →
  the UI, mirroring the SPIN ingress. — `docker-compose.yml`, `nginx.conf`
  (`653a763`, `bd9d88a`)
- **OV deployment files & docs**: Dockerfile, `DEPLOYMENT.md` (incl. how to mint
  OV keys), and an embedding-migration helper. — `knowledge/openviking/*`
  (`4b720aa`, `6b05448`, `c76ff86`)
- **Schema doc**: `ui/app/db/schema.md`

## Design notes
- **OV auth is API-key only** (no passwords). BERIL holds an account-ADMIN key
  for the pre-existing `beril` account; it does **not** bootstrap accounts.
- **Regenerate is opt-in and destructive** — `POST /api/ov/user` never
  auto-regenerates, so a key that may be in use elsewhere is never silently
  invalidated. The regenerate endpoint is scoped to the caller's own ORCiD.
- **Encryption-at-rest caveat**: Fernet protects keys in Postgres dumps/backups,
  but `ov_credential_key` lives in the app env, so an app-server compromise can
  decrypt them. `GET /api/ov/credentials` returns plaintext — serve over HTTPS;
  the key is never logged. Key rotation (MultiFernet) is out of scope.

## Testing
- **11 new route tests** (`ui/tests/test_routes_openviking.py`), run against
  in-memory SQLite with the OV client monkeypatched — cover auth (401), create,
  idempotent repeat, the 409-no-regenerate branch, regenerate/rotate, credential
  fetch (incl. 404), and health (ok/unreachable). `11 passed`.
- **Verified end-to-end against the local docker-compose stack**: identity,
  health, create, and credential-fetch all return 2xx; BERIL authenticates to OV
  with the admin key; the credential row persists with `encrypted_key` as
  ciphertext.

## Deploy / config notes
- Set `BERIL_OV_CREDENTIAL_KEY` (generate with `Fernet.generate_key()`) before
  enabling these endpoints — otherwise create/fetch raise a clear 500.
- `BERIL_OV_ADMIN_KEY` must be the ADMIN key for the OV `ov_account_id` account
  (validate against `GET /api/v1/admin/accounts/<account>/users` → 200).
- Run `alembic upgrade head` to create the `ov_user_credential` table.


briehl:
Note that `nginx.conf` and `docker-compose.yaml` are only used for local development and testing. They reasonably closely mirror what happens on SPIN.

dileep-kishore (approved):
Looks good!

PR #304, opened 2026-06-26
Fix/iceberg migration

mikacashman:
Fixes for iceberg migration (convention of namespaces from `tenant_database` to `tenant.database`.
Changes made to:
- Documentation (including pitfalls detailing changes)
- Hard-coded namespaces
- Methods pulling from inventory


mikacashman:
Added note:  Changes for iceberg migration still TODO for ui and atlas.  To be done in a separate PR due to overlap with OpenViking and requirements to update `collections snapshot` to propagate changes.

scripts/discover_berdl_collections.py
briehl, line 1: Is this a thing we'll just have to update from time to time? I feel like there should be an API call to K-BERDL or something.

But this is fine for now, don't block on it.

briehl (approved):

PR #305, opened 2026-06-30
feat: split research planning from execution, and give the plan-review checkpoint a witness

dileep-kishore:
## Summary

Splits the research arc into **plan → approve → execute**, and gives the plan-review checkpoint the one thing it never had: an artifact.

```
/research-plan  ──►  beril approve  ──►  /execute-plan  ──►  /berdl-review  ──►  /submit
exploration→proposed    (the human)      →active→analysis      (existing)        (existing)
```

`/berdl_start` keeps onboarding and becomes a router keyed on `beril.yaml.status`; the Phase A–D prose moves into the two new skills rather than being duplicated.

## The problem this fixes

`status: proposed` is written by the agent, so it proved nothing. A plan never shown to anyone and a plan a scientist read and accepted were **byte-identical on disk**. `/execute-plan` was told to "verify the user explicitly approved" with nothing to verify against — and its own frontmatter named `status: proposed` as the approved state, which is exactly the *un*approved one. The agent was both the actor and the witness.

## What's new

| Construct | What it does |
|---|---|
| **`/research-plan`** | Interests + live data + literature → a frozen, pre-registered `RESEARCH_PLAN.md`: FINER/PICO question, competing hypotheses with genuine rivals, a falsification test **and** decision criterion each, a discrimination strategy, and a feasibility verdict. A `not-answerable` verdict stops the plan instead of freezing an unanswerable question. Ends at the mandatory checkpoint. |
| **`/execute-plan`** | Reads the frozen plan and runs each notebook's **discriminating query first** (Platt), then `/synthesize`. A material plan change demotes `active → proposed` and re-runs the checkpoint. |
| **`beril approve <id>`** | Records `plan_approval: {by, at, via, plan_hash}` — the one place a human approval becomes a checkable fact. Run it at a terminal (`via: terminal`), or let the agent record it with `--relayed` once you approve in conversation (`via: agent-relayed`). A bare non-TTY call refuses, so relaying is deliberate. |
| **`.claude/hooks/plan-gate.py`** | A `PreToolUse` **witness**. It never blocks. When analysis code is written under a plan carrying no matching approval, it appends a record to `projects/<id>/plan_deviations.jsonl`. |
| **Binding pre-registration** | Each hypothesis states its **unit of analysis** — what one row is, why those rows are independent — before its decision criterion is set. And `/synthesize` now adjudicates **each** hypothesis against **its own** criterion. Previously the plan was frozen and then never checked: `grep -ci "decision criterion\|confidence prior"` on the synthesis skill returned **0**. |
| **`SCOREBOARD.md`** | One row per planned notebook, filled by the agent in the loop that already commits. Progress reads "H2 refuted at NB03 by its stated criterion", not "3 of 5 done". Row count is fixed by the frozen plan; standings use the existing claims vocabulary. |
| **Falsifiability rubric §8** | *"A plan that cannot state what result would refute its hypothesis is not yet testable."* Plus a `hypothesis-critic` agent and `/critique-hypotheses` for pre-analysis stress-testing. |

## Why a witness and not a gate

The first version of this **blocked** unapproved analysis writes. That was wrong: exploration and light analysis are ordinary research and shouldn't need a ceremony first. Pre-registration integrity doesn't require that departing from the plan be impossible — only that it be **undeniable**.

An empirical probe settled the mechanism: `permissionDecision: "ask"` is auto-approved under this repo's `defaultMode: "auto"`, so a hook here can only allow or deny. Recording beats both. The reviewer reads the records under Methodology and `/submit` warns on them; **neither blocks**, so the only hard gate in the repo remains the ORCID gate at submit.

## Design notes

- **The gate never reads `status`.** An agent writing `status: active` unlocks nothing.
- **`plan_hash` covers the plan minus the Revision History *section*.** So logging a minor deviation keeps an approval valid, while editing a hypothesis, a threshold, or the discrimination strategy invalidates it — mechanically enforcing the minor/material rule the repo already had in prose. (Deliberately *not* `tools/review.sh`'s whole-file hash, which answers a different question.)
- **`plan_digest` is duplicated** in the hook and the CLI: the hook runs on system Python without the venv and can't import `beril_cli`. A test pins that the twins agree on output — if they drift, `approve` writes a hash the witness can never match.

## Honest limits

- The witness is **not a security boundary** and isn't meant to be. The digest is computable and the record forgeable, and an agent can relay an approval it was never given — `via: agent-relayed` makes that visible, not impossible. What changes is that a silent, traceless omission becomes a fabricated record sitting in the diff next to the plan it claims to approve. The real control is PR review.
- **Bash writes don't fire the hook** (`Write|Edit|NotebookEdit` and `mcp__.*` only). Accepted: every Bash matcher tried produced false positives on `git diff` and was defeated by one `cd`.
- `plan_deviations.jsonl` is deletable before its first commit, and a project with no plan is never watched.

## Testing

`uv run pytest tests/ -q` → **389 passed**. The hook is exercised as a subprocess under system Python 3.9.6 (what the harness runs it with); `beril approve` round-trips under a real pty on both the terminal and relayed paths.

The suite is deliberately lean — 19 test functions, cut from 45. Every defect found across both review rounds was reintroduced one at a time to confirm the slimmer suite still goes red for each, so the pruning is coverage-preserving rather than coverage-shedding.

Reviewed by Codex (10 findings, all fixed) and by an independent adversarial pass (28 raised, 21 refuted on reproduction, 7 fixed). Both found real defects — a TOCTOU hole where the plan was hashed *before* the confirmation prompt, read-only MCP calls recorded as deviations, paths outside the repo attributed to it, a digest blind spot that 7 existing plans fall into, and a hook wrapper that could return exit 2 (the *block* code) and break every write.

## Not in this PR

The cross-session world-model (`beril state` / `whereami`, `research_state.json`, SessionStart re-injection) was cut to a follow-up — it's a separate feature and was a third of the diff.


briehl (approved):
I went through this in BERDL hub, and it worked as expected. Looks good!

PR #308, opened 2026-07-01
auth: extend UserApiToken for multi-token PATs (1/5)

briehl:
## Summary

First of a 5-PR stack that adds a proper personal-access-token flow so
the `beril` CLI can log a user in from any terminal (including
JupyterHub on K-BERDL) without needing browser cookies.

This PR is **schema + CRUD only** — no route or template changes. The
new fields are wired up but nothing new is exposed to end users yet.

### Model changes to `UserApiToken`

- Drop `UNIQUE(user_id)` → plain index. A user may now hold many tokens.
- Add nullable `name`, `expires_at`, `revoked_at`.
- Legacy rows (from `get_or_create_api_token`) remain valid — `NULL`
  everywhere means "unnamed, never expires, not revoked."

Alembic migration drops the auto-named `user_api_token_user_id_key`
constraint (via `DROP CONSTRAINT IF EXISTS`, safe on any deploy state)
and adds the three columns.

### CRUD changes (`app/db/crud.py`)

New:
- `create_named_api_token(user_id, *, name, expires_in_days=90)` —
  additive, does NOT delete existing tokens.
- `list_api_tokens_for_user(user_id, *, include_revoked=False)` —
  newest first.
- `revoke_api_token(token_id, user_id)` — idempotent.
- `TOKEN_PREFIX = "beril_"` — new tokens are prefixed for grep-ability
  in logs and secret scanners. Prefix is part of the hashed input, so
  validation is transparent — legacy unprefixed 64-hex tokens still
  hash and validate correctly.

Modified:
- `get_user_by_api_token` now rejects revoked and past-`expires_at`
  tokens.
- `_as_aware` helper — SQLite drops tzinfo when reading back
  `DateTime(timezone=True)`; coerce to UTC before comparing to `now()`.

Unchanged:
- `get_or_create_api_token` behaves exactly as before, so the existing
  `POST /api/user/token` endpoint keeps working. Deprecation/removal
  decision deferred to phase 4.

### Test coverage

- 15 new tests in `tests/db/test_crud.py` covering the three new
  helpers, expiry/revocation rejection paths, and confirming legacy
  unprefixed tokens still validate.
- `tests/db/test_models.py` — flipped `test_user_can_have_only_one_token`
  to `test_user_can_have_multiple_tokens`, added a test for the new
  named/expiring/revocable shape.
- Full suite: **757 passing**.

### Stack plan

1. **This PR** — schema + CRUD.
2. Settings UI (`/settings/tokens` list/create/revoke).
3. `/api/user/whoami` route.
4. Bearer-wins auth ordering (small correctness fix in
   `app.auth.get_current_user_session_or_token`).
5. `beril auth login/status/logout` CLI extension.

## Test plan
- [ ] `cd ui && uv run pytest tests/ -q --no-cov` — 757 tests pass.
- [ ] `cd ui && uv run alembic upgrade head` on a dev database runs
      cleanly.
- [ ] `cd ui && uv run alembic downgrade -1 && uv run alembic upgrade head`
      round-trips cleanly (only when there's at most one token per user
      in the DB — the downgrade restores the UNIQUE constraint).

### DIFF
--- a/ui/alembic/versions/20260701_e5f6a1b2c3d4_extend_user_api_token.py
+++ b/ui/alembic/versions/20260701_e5f6a1b2c3d4_extend_user_api_token.py
@@ -1,0 +1,51 @@
+"""Extend user_api_token: multi-token per user, add name/expires_at/revoked_at.
+
+The initial-schema migration declared ``user_id`` UNIQUE, enforcing one token
+per user with rotation-on-create. The new PAT flow needs multiple named,
+expiring, revocable tokens per user, so we drop that unique constraint and
+add columns for name, expiry, and revocation. All new columns are nullable
+so pre-existing rows remain valid: NULL means "unnamed, never expires, not
+revoked."
+
+Revision ID: e5f6a1b2c3d4
+Revises: d4e5f6a1b2c3
+Create Date: 2026-07-01
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+revision = "e5f6a1b2c3d4"
+down_revision = "d4e5f6a1b2c3"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # Drop the auto-named UNIQUE constraint on user_id. PostgreSQL default
+    # name for an inline column UNIQUE is "<table>_<column>_key".
+    op.execute("ALTER TABLE user_api_token DROP CONSTRAINT IF EXISTS user_api_token_user_id_key")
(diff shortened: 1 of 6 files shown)


dileep-kishore (approved):
Looks good. Are we keeping the default token expiration as 90 days?

briehl:
Oops, just set that to a year. Thanks.

PR #311, opened 2026-07-07
docs(onboarding): surface OpenViking setup front and center

dileep-kishore:
## Why

OpenViking is the **default knowledge layer** for BERIL, but its setup was easy to miss: buried in a low-visibility README subsection, and a user who skips the README and just launches the agent got only a one-line pointer with no actual steps. This makes OpenViking onboarding prominent and self-serve.

## What

- **README** — adds a prominent 🧭 callout in the Quick Start flow framing OpenViking as the default knowledge layer with a "do this first" nudge, linking to `docs/remote-openviking-setup.md` as the single source of truth. No step duplication, so the temporary dev→prod host change touches only the doc.
- **`berdl_start` Phase 1.8** — expands the `NO API KEY` / `AUTH FAILED` branch into a condensed **4-step, non-blocking** walkthrough (log in → copy `beril_session` cookie → run `setup_remote_ov.py` in the user's own terminal → verify with `doctor`), so a user who never read the README still gets step-by-step guidance when they launch the agent. Includes a "never paste the cookie into chat" security note.
- **Accuracy fix** — corrected an inherited inaccuracy: on `NO API KEY` / `AUTH FAILED` the server is reachable, so `knowledge_query.py` does **not** fall back to local search (that only applies to `UNREACHABLE`). The guidance now reflects that setting up the key is what unblocks `find`/`grep`/`read`/`overview`.

## Scope

Docs/skill-text only — no logic changes. `OK` / `UNREACHABLE` branches and the non-blocking Phase 1.8 flow are unchanged.

## Notes

- The dev host `beril-dev.kbase.us` is referenced in exactly the two edited spots plus the canonical doc; when OpenViking moves to production, update `docs/remote-openviking-setup.md` (one `SERVER=` line) and the `berdl_start` bullet.
- Reviewed with Codex: both findings (placeholder wording consistency + the fallback-claim bug) were verified against the code and fixed before commit.

### DIFF
--- a/.claude/skills/berdl_start/SKILL.md
+++ b/.claude/skills/berdl_start/SKILL.md
@@ -184,11 +184,21 @@ uv run --env-file .env knowledge/scripts/knowledge_query.py doctor
   **Do not enumerate `projects/` or read `projects/*/README.md` / REPORTs by hand for
   context** — the server has them indexed and ranked, and hand-scanning misses the
   cross-project connections it surfaces.
-- **NO API KEY / AUTH FAILED** → the remote layer needs a key. Tell the user:
-  `uv run knowledge/scripts/setup_remote_ov.py --cookie '<beril_session>'`
-  (walkthrough: `docs/remote-openviking-setup.md`). Until then keep using
-  `knowledge_query.py` — `find` / `grep` / `read` / `overview` fall back to local
-  search over the same corpus.
+- **NO API KEY / AUTH FAILED** → server is up but you have no valid key. OpenViking is
+  the default knowledge layer, so surface these steps to the user, then continue
+  (non-blocking — don't wait on them):
+
+  > **Set up OpenViking (~2 min, one-time):**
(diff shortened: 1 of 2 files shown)


mikacashman:
Should OV setup happen before or after `beril` start?  I might suggest moving the indented paragraph above the `doctor` and `start` commands (just the line above).  There are also now duplicated docs in the readme for OV as well since the separate section is still there.  Probably can just purge the other section, the instructions should link to the other guide anyway since the setup is more involved than running those two uv commands

dileep-kishore:
Thanks @mikacashman. I moved the OpenViking callout above the `beril doctor`/`beril start` line so setup happens first, and purged the duplicate "Knowledge context" section (the callout links to `docs/remote-openviking-setup.md`, which is the right home for the more involved cookie/rotation steps).

mikacashman (approved):
Looks good

cjneely10 (approved):

PR #318, opened 2026-07-08
feat(inventory): cache berdl_start inventory to fix slow startup

dileep-kishore:
## Why

`berdl_start` was slow on every run because Phase 1.6 (`scripts/berdl_inventory.py`) did a **full live Spark fetch every time** — off-cluster this cold-spawns the JupyterHub server; on-cluster it walks the metastore. There was no cache: `data/berdl_inventory.md` is an output file overwritten each run, and the repo's `build_data_cache.py` caches UI/repo content, a separate subsystem.

## What

Cache the raw `structure` + `tenants` to `data/berdl_inventory_cache.json` (gitignored) and serve it **before any Spark import**, so a hit never spawns the server. All output is re-rendered from the cache, so formatting flags keep working on a hit.

- **Auto-refresh past a 7-day TTL** (`--ttl-days N` / `BERDL_INVENTORY_TTL_DAYS` to tune).
- **Keyed by environment + token fingerprint** (sha256 of `KBASE_AUTH_TOKEN`, first 12 hex — the token is never stored). An on/off-cluster or token switch is a miss.
- **Freshness banner** ("Cached 3 days ago … `--refresh` to update") stamped into **both** stdout and the `.md` report — which also fixes a pre-existing gap where a leftover `.md` carried no provenance.
- New flags: `--refresh` (force + rewrite), `--no-cache` (bypass read+write), `--ttl-days N`.
- `berdl_start` SKILL.md Phase 1.6 documents the cache and refresh command.

## Tests

- 24 new tests: token fingerprint, JSON round-trip, corruption→miss, freshness across env/token/TTL boundaries, banner formatting, flags, and every `main()` path (hit/expired/first-run/refresh/no-cache/ttl). An autouse fixture isolates the default cache path so the suite never touches the repo cache.
- Also migrated 21 **pre-existing failing tests** (unrelated to caching) from the legacy underscore tenant model to Iceberg dotted `catalog.namespace` identifiers, which the current source requires. Full file: **56 passed**.
- Verified end-to-end (mocked fetch): first run writes the cache and prints "first run"; the second run prints "Cached … ago" and never calls the fetch.

## Notes

Environment detection (Phase 1.5) is intentionally **not** cached — it must probe connectivity live.


dileep-kishore:
## Testing follow-up: three fixes pushed (1880db74)

Verified the cache end-to-end against the live on-cluster path. The core optimization works — **cold start 258s → warm start 64ms**, with `pyspark`/`berdl_notebook_utils` confirmed absent from `sys.modules` on a hit, and the cached report body byte-identical to the live one. Testing surfaced three defects, now fixed.

### 1. A partial fetch was cached for the full 7-day TTL (the real one)

`berdl_notebook_utils.get_tables()` swallows a per-database listing failure and returns `[]` ("Empty list on lookup failure (logged)"), so a transient `UNAUTHENTICATED` blip on one namespace looks like an *empty database*, not an error. The fetch returned `rc=0` and the lossy snapshot got cached.

This actually happened during testing:

```
cold:     9 tenants · 73 databases · 1405 tables
refresh:  9 tenants · 73 databases · 1404 tables
  | `planetmicrobe.ncbi_taxon_mapping` | 1 | `ncbi_taxon_mapping` |      ← before
  | `planetmicrobe.ncbi_taxon_mapping` | 0 | _(empty or inaccessible)_ |  ← cached
```

A later refresh recovered all 1405 tables, confirming it was transient. That's what makes it worth fixing: **before this PR the blip was visible immediately and self-healed next run; with the cache it gets pinned for a week behind a confident "Cached 3 days ago" banner.**

Fix: watch the WARNING those failures log, mark the fetch degraded, and never cache it — preserving any existing older-but-complete snapshot. The banner reads `partial … not cached` so undercounted totals aren't mistaken for truth.

Degradation is deliberately scoped to *database-listing* failures via a dedicated `berdl_inventory.structure` logger. A failed `get_tenant_detail` costs a display name, not a database; blocking the cache on it would let one permanently-403ing tenant disable caching forever for that user and restore the slow startup this PR exists to remove.

### 2. A malformed `BERDL_INVENTORY_TTL_DAYS` crashed the script

`int("")` → `ValueError`. Since `berdl_start` shells out to this script on every startup, an exported-but-empty var replaced the entire inventory with a traceback. Now falls back to the default with a warning.

### 3. A tz-naive `fetched_at` in the cache crashed `_cache_is_fresh`

It parses cleanly via `fromisoformat`, then raises `TypeError` subtracting from the aware `_now()` — breaking `load_cache`'s documented "any failure is a miss, never a crash" contract. Now treated as malformed.

### Also

`--refresh` and `--no-cache` no longer mislabel themselves as "cache expired". The cache was deliberately bypassed, not stale.

### Tests

21 new tests, **77 passing** in `tests/scripts/test_berdl_inventory.py` (89 across `tests/scripts`). Covers: the degraded-fetch guard (not cached, existing cache preserved, clean fetch still cached), the structure-vs-tenant logger split, malformed/empty TTL env vars, the naive-timestamp miss, and the corrected banners.

I also confirmed the 21 pre-existing failures this PR migrates do fail on `main` — that claim checks out.

mikacashman:
@dileep-kishore do you approve of the short addition I committed?

dileep-kishore:
Thanks, those changes are helpful! This should speedup the 4m+ inventory times I am seeing

mikacashman (approved):
Look good, startup time improved across the board

PR #331, opened 2026-07-20
cli: beril auth login/status/logout for CLI authentication

briehl:
## Summary

Final phase of the PAT flow. Adds `beril auth login/status/logout` so
agents running from a terminal — locally, on JupyterHub, wherever — can
hold a BERIL server credential and act on the user's behalf without a
browser session.

### Commands

```
beril login [--token TOKEN] [--base-url URL] [--status]
```
1. If `--base-url` given, use and persist it under `[beril].base_url`
   in `~/.config/beril/config.toml` so future runs don't need it.
2. If `--token` given, use it directly (scriptable / headless).
   Otherwise print instructions ("open `<base>/account/tokens`, create
   a token, paste it") and read via `getpass` so it does not echo.
3. Validate by calling `GET /api/user/whoami` with
   `Authorization: Bearer <token>`. On 2xx, persist token + resolved
   ORCiD + display name to `~/.beril/auth.json` at mode `0600`. On
   failure, print a specific error and exit 1 without writing anything.
4. If `--status` given, do not update the token, just check the current login status and validate the existing token, if one is there.

```
beril logout
```
Removes `~/.beril/auth.json`. Idempotent (still exits 0 when already
logged out, but says so).

### Design notes

- **`httpx` for HTTP.** urllib was the first attempt (to keep the CLI
  dep-free) but the prod deployment sits behind Cloudflare, which
  rejects the default `Python-urllib/*` User-Agent with a 403. httpx's
  default UA sails through. Also added `certifi` so a freshly-installed
  python.org Python does not fail cert verification against the prod
  cert chain.
- **Mode 0600 at file-creation time.** `auth_store.save` uses
  `os.open(..., O_CREAT | O_WRONLY | O_TRUNC, 0o600)` so the file is
  never briefly world-readable — a `write_text` + `chmod` sequence
  would race with any process that stat's the file between the two
  calls.
- **Base URL resolution order**: `BERIL_BASE_URL` env var →
  `[beril].base_url` in config → compiled-in default
  (`https://beril.kbase.us`). Trailing slashes stripped consistently so
  path concatenation is safe.
- **Distinct error messages** for 401, 403, generic 4xx/5xx, timeout,
  transport error, invalid-JSON, and missing-`orcid_id` responses — so
  a failing `login` tells the user exactly which layer said no.

Verified end-to-end against `https://beril.kbase.us`.

## Test plan

- [ ] `uv run pytest tests/ -q` from repo root — 302 pass.
- [ ] `beril login --base-url https://beril.kbase.us`, paste a
      token from `/account/tokens`, confirm `beril auth status` shows
      the right identity.
- [ ] `curl -H "Authorization: Bearer $(python -c 'import json,pathlib;
      print(json.loads(pathlib.Path.home().joinpath(".beril/auth.json").read_text())["token"])')"
      https://beril.kbase.us/api/user/whoami` — returns the same
      identity.
- [ ] `beril logout` — file gone, `beril auth status` exits 1.
- [ ] `beril login --token X` (bad token) — prints 401 error,
      does not overwrite an existing auth.json.

### DIFF
--- a/.claude/skills/knowledge-context/SKILL.md
+++ b/.claude/skills/knowledge-context/SKILL.md
@@ -19,8 +19,12 @@ OpenViking holds an indexed context layer over BERIL projects and central docs. 
 ## Picking the right primitive
 
 ```
-QUERY="uv run --env-file .env knowledge/scripts/knowledge_query.py"
+QUERY="uv run knowledge/scripts/knowledge_query.py"
 ```
+
+Credentials resolve from `~/.beril` after `beril login` (see § Configuration) —
+no `--env-file` needed. To override, set `OPENVIKING_URL` / `OPENVIKING_API_KEY`
+in the env or prepend `--env-file .env`.
 
 | Question | Use | Why |
 |---|---|---|
@@ -216,7 +220,7 @@ For everything else (project→doc, doc→doc, file-level links), use the `link`
 ## Refreshing context
 
 ```bash
-INGEST="uv run --env-file .env knowledge/scripts/ingest_context.py"
+INGEST="uv run knowledge/scripts/ingest_context.py"   # creds from ~/.beril; see § Configuration
 $INGEST --project <project_id>     # after editing one project
 $INGEST --changed                  # after mixed edits — diffs against state manifest
 $INGEST --all                      # full refresh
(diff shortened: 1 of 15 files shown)


dileep-kishore (approved):
I think urllib gets 403'd for sending no custom UA rather than for being urllib specifically? If that's right, maybe we should just set an explicit UA so we're not leaning on httpx's default staying off Cloudflare's list?

briehl:
> I think urllib gets 403'd for sending no custom UA rather than for being urllib specifically? If that's right, maybe we should just set an explicit UA so we're not leaning on httpx's default staying off Cloudflare's list?

It'll be the same issue. If Cloudflare decides to add whatever explicit UA we pick to their disallowed list, we'll be in the same boat. And it's generally not a great practice to spoof UA if we can help it.

I'll add a note to some developer documentation.

PR #204, opened 2026-04-24
Add two pitfalls from obi_ontology_coverage

turbomam:
Documents two reproducible issues discovered during `obi_ontology_coverage` work:

1. **Parquet `fixed_size_binary[16]` UUID error** — Spark 4.x rejects 16-byte fixed binary columns (e.g. MD5 hashes) exported by DuckDB. Includes pre-screen + conversion workaround.
2. **`/berdl-ingest` on-cluster failure** — `initialize()` checks for SSH tunnels on ports 1337/1338, which don't exist on JupyterHub. Documents the on-cluster workaround and the fix needed (`SPARK_CONNECT_MODE_ENABLED` detection).

PR #332, opened 2026-07-21
Add obi_ontology_coverage ingest-session docs + two general lakehouse pitfalls

turbomam:
Rebased replacement for #204, per Bill's request there to update from main and get a fresh Copilot review. #204's head was my fork's `main` branch itself; this uses a proper feature branch instead.

Verified while rebasing: the `/berdl-ingest` on-cluster limitation this documents is still unfixed on current main (`initialize()` in `scripts/ingest_lib.py` has no on-cluster/`SPARK_CONNECT_MODE_ENABLED` detection), so this pitfall entry is still accurate and nothing went stale.

Original description:
- Parquet `fixed_size_binary[16]` causes a Spark UUID error (DuckDB exports hash columns this way; pre-screen with PyArrow and convert to hex strings)
- `/berdl-ingest` skill doesn't support on-cluster execution (assumes SSH tunnels that don't exist when running on the notebook server itself)


turbomam:
this was a fun way to learn about BERIL etc but isn't' mission critical and not worth maintaining in my eyes.

PR #333, opened 2026-07-21
Give beril --version a meaningful, git-derived output

### LINKED ISSUE #309: give the `beril --version` a meaningful output
it has been 0.1.0 for a while

turbomam:
Closes #309.

`version` had been a static, hand-maintained duplicate string in `pyproject.toml` and `beril_cli/__init__.py`, stuck at "0.1.0" while real release tags moved on through v0.3.4.5. Switched to `hatch-vcs` so the version is derived from git tags/commits at build time, read at runtime via `importlib.metadata` instead of a hardcoded string.

Verified: `pip install -e .` in a clean venv, then `beril --version` reports `0.0.6.dev116+ga8c7a516c.d20260721` instead of the stale `0.1.0`. Ran the CLI test suite (`tests/test_cli_*.py`, 49 tests): all pass. Unrelated to this change, the repo's other test suites (spark/project-specific) need dependencies I didn't install, so I scoped verification to what this change touches.

### DIFF
--- a/beril_cli/__init__.py
+++ b/beril_cli/__init__.py
@@ -1,1 +1,8 @@
-__version__ = "0.0.6"
+from importlib.metadata import PackageNotFoundError, version
+
+try:
+    __version__ = version("beril-cli")
+except PackageNotFoundError:
+    # Running from source without an installed distribution (e.g. editable
+    # checkout that was never `pip install`-ed) — no version metadata exists.
+    __version__ = "0+unknown"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "beril-cli"
-version = "0.0.6"
+dynamic = ["version"]
 description = "BERIL Research Observatory — CLI launcher and environment manager"
 requires-python = ">=3.11"
 license = "MIT"
@@ -23,5 +23,8 @@ knowledge = [
 ]
 
 [build-system]
-requires = ["hatchling"]
+requires = ["hatchling", "hatch-vcs"]
 build-backend = "hatchling.build"
+
+[tool.hatch.version]
+source = "vcs"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,3 +28,7 @@ build-backend = "hatchling.build"
 
 [tool.hatch.version]
 source = "vcs"
+# Only three-component release tags are eligible. Without this, any tag becomes a
+# version candidate: the repo already carries `archive/verify-ingest-2026-04-attempt`,
+# which `git describe` cannot parse, and `v0.3.4.5`, which is not our scheme.
+raw-options = { tag_regex = '^v(?P<version>[0-9]+\.[0-9]+\.[0-9]+)$' }
--- a/tests/test_cli_version.py
+++ b/tests/test_cli_version.py
@@ -1,0 +1,20 @@
+"""Tests for the installed CLI distribution version."""
+
+import importlib.metadata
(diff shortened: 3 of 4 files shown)


turbomam:
This may not address the fact that there are multiple components in this ecosystem and that they might legitimately have different versions, or that synchronizing version might take some maintenance effort

turbomam:
Somewhat related to @briehl's 
- https://github.com/kbaseincubator/BERIL-research-observatory/issues/230 

This PR moves versioning to git-tag-derived (hatch-vcs), which would give that publish-on-merge flow an automatic version on every build instead of the hand-maintained string it replaces. Flagging it here so the versioning scheme can align with whatever publish workflow you have in mind.

turbomam:
Rebased onto `main`. Conflicts were stale manual version bumps; the diff stays scoped to `beril-cli`. `beril --version` now reports a git-derived version rather than the hard-coded `0.0.6`. CI green.

turbomam:
Follow-up after adding the focused fallback-version coverage test: all three CI workflows pass, and GitHub reports commit a50eff98 cleanly mergeable.

dileep-kishore:
@turbomam The hatch-vcs change makes sense to me. Should the intended workflow be to create a GitHub Release tag, then have CI build the package and verify that beril --version matches it? If so, should hatch-vcs only accept standard release tags like vX.Y.Z?

turbomam:
Added tests for the metadata-lookup success path in 0f0db852, covering the suppressed comment in review 4964559021 and closing https://github.com/kbaseincubator/BERIL-research-observatory/issues/395 . Verified they fail if the distribution name is wrong.

turbomam:
Correction to the commit message on 0f0db852: it cites "review 4993931118". That is not a review on this pull request and I should not have written a number I had not looked up. The two reviews here are 4746133448 (2026-07-21, no suppressed block) and **4964559021** (2026-08-18), and 4964559021 is the one carrying the suppressed comment this commit addresses, as stated in my previous comment.

The commit message is wrong; the acknowledgement above is right. Not amending, because that would mean force-pushing this branch.

turbomam:
@dileep-kishore yes to both, and done in 87b51ba6.

`hatch-vcs` now only accepts `vX.Y.Z`. Worth doing because the repo already has two tags that break it: `archive/verify-ingest-2026-04-attempt`, which `git describe` cannot parse, and `v0.3.4.5`.

Release tag, CI builds, CI asserts `beril --version` matches is the workflow I would want too.

One thing I cannot decide: `v0.3.4.5` is reachable from `main` but further away than `v0.1.1`, so `git describe` picks `v0.1.1` and a build derives `0.1.2.dev...`, which is lower than a tag that already exists. Was `v0.3.4.5` a real release, or a mistake we should delete? I have not touched any tags.

dileep-kishore:
I think `v0.3.4.5` was accidentally added by a project when it was merged.
@briehl does this look good to you?

turbomam:
The suppressed finding from the 2026-08-18 review is covered. `tests/test_cli_version.py` now has a success-path test, a test asserting the queried distribution name matches the one declared in `pyproject.toml`, and one unmocked lookup against the real environment. The middle one guards the exact case the review raised: a wrong name would report `0+unknown` while the fallback test still passed.

briehl:
I removed the `v0.3.4.5` tag, so that shouldn't be an issue anymore.
I think this is fine to merge. 👍

briehl (approved):

PR #338, opened 2026-07-22
beril login: link and cache the OpenViking credential

briehl:
## What

Makes `beril login` the **single call** that links a user to both BERIL and OpenViking (OV). After the token is validated, login now provisions the user's OV account (if needed), fetches the OV credential, and caches it alongside the BERIL auth token in `~/.beril/auth.json`. No browser cookie copy, no manual `.env` editing.

This is the first of a stacked series (PR2: `beril ov` repair/rotation command; PR3: query-CLI config fallback + docs).

## Changes

- **`auth_store.py`** — `AuthRecord` gains optional `ov_url` / `ov_user_key`; `save()` persists them; `load()` reads them with `.get()` so **pre-OV `auth.json` files still parse** (read as logged-in / OV-unlinked). New `load_ov()` returns `(url, key)` or `None`.
- **`ov_client.py`** (new) — shared Bearer-authed helper: POST `/api/ov/user` (or `/api/ov/user/regenerate`) then GET `/api/ov/credentials`. Raises `OvLinkError`, flagging the recoverable **409** "OV user exists but BERIL holds no key" case (`needs_regenerate`). This is the headless replacement for the browser-cookie flow in `knowledge/scripts/setup_remote_ov.py`, and is shared so the upcoming `beril ov` command reuses the same endpoint/error logic.
- **`auth_cmd.py`** — `run_login` links OV after `whoami` succeeds.

## Key design decision: OV linking is non-fatal to login

A valid BERIL login must still succeed even if OV is unreachable or returns the 409 case. In those cases login saves the BERIL token with OV fields `None`, exits 0, and prints guidance to run `beril ov setup [--regenerate]` (that command lands in PR2). Only a hard BERIL auth failure is non-zero (unchanged).

## Tests

- `auth_store` OV round-trip, defaults-to-None, old-file-without-OV-fields still loads, `load_ov()` cases.
- `run_login`: caches OV cred + masks the key in output; OV-unreachable still succeeds (exit 0, fields None, warning); 409 points at `--regenerate`.
- `ov_client`: happy path, Bearer header, regenerate endpoint, 409→`needs_regenerate`, 401, missing key, transport error, trailing-slash handling (via `httpx.MockTransport`).

Full suite green (321 passed).

### DIFF
--- a/.claude/skills/knowledge-context/SKILL.md
+++ b/.claude/skills/knowledge-context/SKILL.md
@@ -19,8 +19,12 @@ OpenViking holds an indexed context layer over BERIL projects and central docs. 
 ## Picking the right primitive
 
 ```
-QUERY="uv run --env-file .env knowledge/scripts/knowledge_query.py"
+QUERY="uv run knowledge/scripts/knowledge_query.py"
 ```
+
+Credentials resolve from `~/.beril` after `beril login` (see § Configuration) —
+no `--env-file` needed. To override, set `OPENVIKING_URL` / `OPENVIKING_API_KEY`
+in the env or prepend `--env-file .env`.
 
 | Question | Use | Why |
 |---|---|---|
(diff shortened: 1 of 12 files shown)


briehl:
Went over this in the standup today and decided to merge.

PR #339, opened 2026-07-22
beril ov: link/inspect/export the OpenViking credential

briehl:
## What

Adds the `beril ov` command — the OpenViking (OV) linking cases `beril login` intentionally doesn't force. All subcommands reuse the stored BERIL PAT (`~/.beril/auth.json`) and the shared `ov_client` helper from PR1 — **no browser cookie**.

```
beril ov setup                re-link OV against the stored login (repair path:
                              login couldn't reach OV, or an old auth.json
                              predates OV linking)
beril ov setup --regenerate   mint a fresh OV key (invalidates the old one) —
                              recovery for the 409 case, and key rotation
beril ov status               show the cached credential (masked) + probe
                              /api/ov/health
beril ov print-env            emit OPENVIKING_URL / OPENVIKING_API_KEY lines
                              for .env or eval "$(beril ov print-env)"
```

> **Stacked on #338** (`beril login: link and cache the OpenViking credential`). Base branch is `pr1-login-links-ov`; review/merge that first. The diff here is only the `beril ov` commit.

## Changes

- **`ov_cmd.py`** (new) — the command. `setup` re-persists the *full* auth record, so the BERIL token/identity are preserved and only the OV fields change. `status`' health probe is best-effort — a probe failure prints `unknown (...)` and still exits 0.
- **`ov_client.py`** — add `ov_health()` so the `/api/ov/health` HTTP logic lives with the rest of the OV client.
- **`cli.py`** — `ov` subparser with required `setup` / `status` / `print-env` subcommands + dispatch.

## Exit codes

- `beril ov` with no subcommand → 2 (argparse usage).
- any subcommand with no stored login → 1 ("Not logged in").
- `status` / `print-env` when OV isn't linked → 1 with a "not linked" hint.
- `setup` on OV failure → 1; the 409 case additionally prints `--regenerate` guidance.

## Tests

`test_cli_ov_cmd.py` (new): setup links+caches / preserves BERIL identity / regenerate flag / 409 guidance / generic failure; status masked-key + health / survives health failure / unlinked; print-env valid lines / unlinked. Plus `ov_health` happy + 401 in `test_cli_ov_client.py`. Full suite green (334 passed).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #340, opened 2026-07-22
knowledge query: resolve OV credential from ~/.beril; docs

briehl:
## What

Closes the loop: the knowledge query CLI now works with **no `--env-file`** once a user has run `beril login`. `ContextConfig.from_env` falls back to the OpenViking credential cached in `~/.beril` when the env vars aren't set.

> **Stacked on #339 → #338.** Base branch is `pr2-beril-ov-command`; review/merge those first. The diff here is only the config-fallback + docs commit.

## Changes

- **`observatory_context/config.py`** — `from_env` precedence:
  1. explicit env vars (`OPENVIKING_URL` / `OPENVIKING_API_KEY`) — always win (CI, `--env-file`, "point at another server"),
  2. the `~/.beril` cached credential via `beril_cli.auth_store.load_ov()`,
  3. the local default URL.

  The `beril_cli` import is guarded (`try/except ImportError`), so `observatory_context` still works in any environment where `beril_cli` isn't on the path — it just behaves as "no cached credential". (In the `uv run --script` query CLI, the repo root is on `sys.path` and `auth_store` is stdlib-only, so the fallback resolves there too.)

- **`knowledge/tests/test_config.py`** (new) — env-wins, cache-fallback, default-URL, partial-env completion from cache, and the import-guard behavior.

- **`.claude/skills/knowledge-context/SKILL.md`** — Configuration section now leads with `beril login` / `beril ov setup` as the primary setup and documents the `from_env` precedence; cookie flow demoted to a fallback mention.

- **`docs/remote-openviking-setup.md`** — rewritten around `beril login`; the browser-cookie walkthrough is **removed**.

## Tests

`knowledge/tests/test_config.py` green (7 tests); full `knowledge/tests` suite green under `uv run --group knowledge` (35 passed). The `beril_cli` CLI suite remains green (334). Verified end-to-end locally: with env vars unset, `ContextConfig.from_env()` resolves the URL + key from a real `~/.beril/auth.json`.

### DIFF
--- a/.claude/skills/knowledge-context/SKILL.md
+++ b/.claude/skills/knowledge-context/SKILL.md
@@ -19,8 +19,12 @@ OpenViking holds an indexed context layer over BERIL projects and central docs. 
 ## Picking the right primitive
 
 ```
-QUERY="uv run --env-file .env knowledge/scripts/knowledge_query.py"
+QUERY="uv run knowledge/scripts/knowledge_query.py"
 ```
+
+Credentials resolve from `~/.beril` after `beril login` (see § Configuration) —
+no `--env-file` needed. To override, set `OPENVIKING_URL` / `OPENVIKING_API_KEY`
+in the env or prepend `--env-file .env`.
 
 | Question | Use | Why |
 |---|---|---|
@@ -216,7 +220,7 @@ For everything else (project→doc, doc→doc, file-level links), use the `link`
 ## Refreshing context
 
 ```bash
-INGEST="uv run --env-file .env knowledge/scripts/ingest_context.py"
+INGEST="uv run knowledge/scripts/ingest_context.py"   # creds from ~/.beril; see § Configuration
 $INGEST --project <project_id>     # after editing one project
 $INGEST --changed                  # after mixed edits — diffs against state manifest
 $INGEST --all                      # full refresh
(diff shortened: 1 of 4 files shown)


.claude/skills/knowledge-context/SKILL.md
```diff
@@ -19,9 +19,13 @@ OpenViking holds an indexed context layer over BERIL projects and central docs.
 ## Picking the right primitive
 
 ```
-QUERY="uv run --env-file .env knowledge/scripts/knowledge_query.py"
+QUERY="uv run knowledge/scripts/knowledge_query.py"
 ```
 
+Credentials resolve from `~/.beril` after `beril login` (see § Configuration) —
```
dileep-kishore, line 25: Are we only planning to store the server/OV credentials in ~/.beril? What about KBase token, ORCID, etc? Will they still be stored in .env?

briehl: That's a good question. I did it in ~/.beril directory to keep it removed from the repo in case that gets wiped or changed for some reason.

But I agree, it should be kept in the same place. Even if the token gets lost, it's pretty trivial to make a new one. I can move that to .env for consistency.

PR #341, opened 2026-07-27
Add a dev mode to the beril start command

briehl:
Right now running `beril start` will pull some version of the codebase (defaulting on the latest release), which might override some local changes.

This adds a `--dev` flag that skips the version check and makes no changes.

Run with `beril start --dev` to make that work.


dileep-kishore (approved):
Looks good. Have we reconciled the git tag versioning with the beril cli versioning yet?

briehl:
> Looks good. Have we reconciled the git tag versioning with the beril cli versioning yet?

Not yet. We had a discussion in last Friday's BERIL meeting but wanted your input before deciding. I think most of us were ok with just going by a single release version for simplicity, but are open to more discussion.

PR #342, opened 2026-07-29
feat: live dashboard, worklog, and status line for in-progress projects

dileep-kishore:
The agent's output scrolls past faster than anyone can follow, and the research decisions are buried in it. This adds three things so you can see what the co-scientist is doing while it works.

## What you get

**`WORKLOG.md`** — an append-only narrative the agent writes as it goes, one entry per lifecycle transition and per unit of work, saying *why* and linking the artifacts that step produced. A leading `!` marks a correction — a bug found, a re-run, an approach abandoned. Written by the `worklog-capture` skill, hooked into `/berdl_start`, `/research-plan`, `/execute-plan`, `/synthesize`, `/berdl-review` and `/submit`.

**A live dashboard** — a page you leave open in a browser tab: research question, hypothesis and what would settle it, the lifecycle stage, the worklog, and every artifact on disk. Refreshes every 4s. Figures open in a lightbox, notebooks and CSVs in JupyterLab's real viewers, markdown in an in-page overlay.

**A status line** — the project, its stage and the dashboard URL, next to a context-usage bar.

## Using it

Nothing to run. The status line starts the dashboard on the first turn a project resolves, and a `SessionEnd` hook stops it when the session ends. On BERDL it's reachable through `jupyter-server-proxy`'s zero-config `/proxy/<port>/` route, inheriting the hub's auth — no config, no Jupyter restart.

It works on the 78 projects already on disk: 61 have no `beril.yaml`, so the stage is inferred from the filesystem and the page says it inferred it.

## Worth a reviewer's attention

**The page refuses to assert what it can't back.** Inferred stage says so; a review whose hash no longer matches `REPORT.md` reads `stale`; an unresolvable worklog link renders struck through; relative times are client-side so a `304` can't freeze a green "alive" dot; notebook progress is a count, never "3 of 5" — plan section numbers don't map to filenames.

**Artifacts come from a filesystem scan, not from `WORKLOG.md`.** The worklog is agent-written prose under a never-block rule and will sometimes be incomplete; a monitoring page can't treat the agent's self-report as ground truth.

**Stdlib-only at runtime**, enforced by a test that parses the module with `ast` — the pod launches it with a bare `python3` and no install.

**The status line has a side effect**, which is unusual for a display component. It's bounded: it only starts a dashboard when a project resolved *and* the port is closed. The launcher used to live in skill prose and never fired during exploration, which is most of a project's early life.

## Testing

**415 passed**, ~6s. CI previously only ran `ui/`, so nothing under the repo's own `tests/` had ever run on a PR; that's fixed here.

Design rationale, including what was tried and rejected (SSE, WebSockets, a pidfile): `docs/live-dashboard-design.md`.


dileep-kishore:
## Merging this alongside #305

Simulated the full sequence against current tips of both branches — merge #305 into `main`, then this PR. **No compatibility issues.** Merged tree: **423 passed, 0 failed, 0 skipped** (the skip disappears because the digest twin test activates).

Order doesn't matter. Whoever merges **second** does the resolution below.

### The resolution — 4 steps

| # | Step | Flagged by git? |
|---|---|---|
| 1 | `.claude/skills/berdl_start/SKILL.md` → take #305's version | yes — conflict |
| 2 | `PROJECT.md` → keep #305's `beril.yaml` line **and** this PR's `WORKLOG.md` line | yes — conflict |
| 3 | Move the worklog + dashboard hooks into `research-plan` / `execute-plan`, and update `worklog-capture`'s transition table | **no** |
| 4 | **Re-add the Phase 0 worklog hook to `berdl_start`** — step 1 deletes it | **no** |

Steps 3 and 4 are silent to git but not to CI: `tests/test_skill_wiring.py` fails naming the exact missing hook. I hit step 4 by accident during the simulation and the test caught it, so this can't be got wrong quietly — it just stays red until finished.

### Verified post-merge

| Check | Result |
|---|---|
| `plan_digest` vs `beril_cli.approve_cmd`, all 73 real plans | 0 diverge |
| Real `beril approve --relayed` → approval chip | `approved`, ORCID + timestamp correct |
| Material plan edit | → `stale` |
| Revision-History append | → stays `approved` |
| Pre-#305 projects falsely accused | `71 na, 7 missing` — not all 78 |
| All 78 projects render | 78 / 0 failed |

### Bugs found and fixed while checking this

- `plan_digest` had drifted from the canonical rule and diverged on **53 of 73** real plans — every approved plan would have rendered `approval stale`. The twin test couldn't catch it: all four fixtures had `## Revision History` last, exactly where the two rules coincide. Fixed and pinned in `b0cf4e7a`.
- Gating the "plan not approved" chip on repo-wide machinery only deferred a false accusation: the day `plan-gate.py` lands, all 78 projects would have flipped to `plan not approved ✗`, 17 of them already `complete`. Same commit.
- `beril approve` records `via: agent-relayed`; the dashboard dropped it and rendered relayed and witnessed approvals identically. Fixed in `ad7c3a2d`.
- The root test suite had **never run on a PR** — the only test step `cd ui`s into a suite scoped to `ui/tests`. Fixed in `09eb6d3a`; that also means this table is the first CI-visible run.

### Follow-ups, none blocking

- **`analysis` → `synthesized`.** The status name means the opposite of what it says — analysis happens during `active`; `analysis` means the report is drafted. ~9 files, zero migration risk (0 of 78 projects sit at that status).
- **Decision criteria vs outcomes on the dashboard.** Needs #305's per-hypothesis table. Not buildable yet: the table went from 4 to 5 columns on 2026-07-29 and no plan on disk has it.
- **Have `/research-plan` name notebooks by filename.** Plan section numbers don't correspond to filenames (measured: one plan's "Notebook 1" is `00_inventory_audit.ipynb`, another's is `02_essential_families.ipynb`), which is why notebook progress here is a count rather than a per-notebook tick. 42/73 plans already name them informally in backticks.

dileep-kishore:
## Three defects from an end-to-end run (`b06238e1`)

Exercised the full `/berdl_start` → plan → checkpoint → analysis → `/synthesize` arc on a real project against live BERDL. The honesty rules all held — inferred stage, stale review, struck-through missing link, client-side relative times — and the missing-link chip even resolved itself back to a live link when the file it pointed at was later created. Three things did not.

**§ Documents could not see `PLAN_REVIEW_N.md`.** `glob("REVIEW*.md")` is anchored at the start of the filename. The artifact `review.sh --type plan` writes — the one this branch's own plan-review checkpoint option (b) tells the operator to produce — was invisible unless a worklog entry happened to link it.

Fixing the glob alone would have been worse than the bug. `review.sh:130` writes `sha256sum` of the **entire** `RESEARCH_PLAN.md` into its `plan_hash` footer, while `plan_digest()` deliberately excludes `## Revision History` so a revision bump doesn't void a human's approval. They're different hashes for the same file — verified `ba53f320…` vs `e64d4e73…` — so chipping plan reviews against the digest would have marked every one of them stale forever. Each family is now checked against its own subject, with a test pinning the divergence so it can't drift back.

**Figures inlined in `REPORT.md` were hidden by CSS.** `.lightbox-overlay.mode-doc img` is a descendant selector: it hid the overlay's own figure `` as intended, and every image embedded in the rendered document along with it. Inlining figures next to the finding they support is the documented contract in the synthesize skill, so reports rendered with none of them. The overlay's `` is a direct child of `#lightbox` and document images are nested in `.lightbox-doc`, so `>` separates them. The test pins that structural assumption — reorder the markup and the selector silently stops hiding the overlay image instead.

**The notebook tally read as a fraction.** `3 planned · 4 written` puts the two numbers in an n/m relationship, so a correct tally looks like an error the moment an unplanned notebook exists — and one always does, because Phase A mandates an exploration notebook the `RESEARCH_PLAN` template never lists. `plan_summary`'s docstring already anticipated this ("unplanned ones appear"); only the presentation implied otherwise. Now `4 written · 4 executed · plan lists 3` — state on disk leads, the plan follows as a labelled reference. No counts hidden, no logic changed.

Each of the four new tests was verified red against the pre-fix code and green after.

### Not fixed here

The statusline's second line never rendered once across the whole session. It's not a display bug — the resolution chain has no reachable signal for a project created *during* the session: cwd predates the directory, the `projects/<id>` branch is optional in Phase 0 step 6, `/add-dir` is manual, and the only writer of `runtime.json` is the `SessionStart` hook, which fires before Phase 0 creates the project. `grep -c runtime .claude/skills/berdl_start/SKILL.md` → `0`.

Worth noting the file *did* eventually appear — written by the `claude -p` subprocesses `review.sh` spawns, each a fresh session firing the hook — but keyed to two dead subprocess IDs, so `from_runtime()` still returns `None` for the real session. Left alone deliberately; it wants a decision about which component owns the post-scaffold snapshot rather than a patch here.

briehl:
I'm having trouble running this on BERDL hub. Should it be runnable there? Or just locally?

briehl (approved):

PR #347, opened 2026-07-30
Ensuring OV context fallbacks

briehl:
This adds a GHA workflow to test the knowledge layer, specifically the code around the OV scripts.

TODO:
* add fallback to search Github project, then Lakehouse projects for context.


dileep-kishore (approved):
Looks good. Seems like the OV commands mostly cleanly map to local versions. I wonder how it was still working fine without these. I guess claude just ignored the knowledge scripts.

PR #351, opened 2026-07-31
feat(dashboard): show when the agent is waiting on you

dileep-kishore:
The dashboard now shows when the agent is blocked waiting on a human, and what it is waiting for.

## Added

- **State chip in the header** — `waiting for you` / `turn ended` / `state unknown`, beside the existing live/idle/cold dot.
- **Amber strip** naming the tool and argument the agent is blocked on (`Bash: sw_vers`, `AskUserQuestion: which cohort?`). Pulses once on the transition, honours `prefers-reduced-motion`.
- **`<title>` marker and favicon dot**, so a background tab shows it without being read.
- **Opt-in system notifications** behind an "Enable alerts" button.
- **`.claude/hooks/agent_state.py`** — writes `projects/<id>/.agent-state.json` on `PermissionRequest`, `Notification`, `Stop` and `UserPromptSubmit`; `SessionEnd` clears it via `dash_stop.py`. Gitignored.
- **A hidden tab now polls** at 15s instead of skipping the fetch entirely — without it none of the above can update while you are away.

## Behaviour worth knowing

`PermissionRequest`, not `Notification`, supplies the detail: the notification payload is a fixed `"Claude needs your permission"` string with no tool in it, and it arrives ~6s later. `PermissionRequest` stays silent for auto-allowed calls, so it only fires when a human is really being asked.

Notifications stay quiet by default. `waiting` always notifies; `turn_ended` only when the tab has been hidden over 60s, since `Stop` fires at the end of every turn. Everything is keyed on `(state, since)` so a re-render never re-notifies.

A `waiting` claim expires: `unknown` after 30 minutes, or as soon as `runtime.json` has no record of the session that wrote it. Otherwise a killed session would leave the page saying "waiting for you" forever.

Detail text is agent-authored, so it goes through the same `inline_md`/`e()` escaping as the worklog.

## Limits

Snapshot mode has no poll — it reports the state it was written with and updates on reload. A closed tab gets nothing; that needs a service worker and a push service. Browsers throttle hidden-tab timers to ~1/min after ~5 minutes, so expect up to a minute of latency when away.

## Tests

444 pass, 16 new. Notification and pulse logic runs under node against a DOM stub (skipped if node is absent) rather than being asserted against the source; the page was also checked in headless Chrome. Design notes in `docs/live-dashboard-design.md`.

### DIFF
--- a/.claude/hooks/agent_state.py
+++ b/.claude/hooks/agent_state.py
@@ -1,0 +1,203 @@
+#!/usr/bin/env python3
+"""Record whether the agent is blocked on a human, for the live dashboard.
+
+Writes `projects/<id>/.agent-state.json`. One dashboard, one project, one file;
+`tools/dashboard.py` reads it in `scan()` and renders a chip, a strip, a title
+marker and a favicon dot from it. Nothing else consumes it.
+
+Registered on four events, and **which four is a measured result, not the
+documented shape**. The hook docs list a `Notification` matcher over
+`permission_prompt | agent_needs_input | idle_prompt | agent_completed` and give
+no payload schema, so this was derived by registering a logging hook against
+Claude Code 2.1.220 and driving a real session through a pty:
+
+- `PermissionRequest` fires ~2s after the blocking tool call and carries
+  `tool_name` + the full `tool_input`. It does **not** fire for a call the
+  permission rules auto-allow (verified by allowlisting `Bash(sw_vers:*)` and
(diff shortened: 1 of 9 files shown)


dileep-kishore:
Closed automatically when the branch was renamed `dileep-kishore/agent-notifications` -> `feat/agent-notifications`. GitHub will not reopen a PR whose head ref name no longer exists. Continues as #354 with the same commits.

PR #354, opened 2026-07-31
feat(dashboard): show when the agent is waiting on you

dileep-kishore:
The dashboard now shows when the agent is blocked waiting on a human, and what it is waiting for.

## Added

- **State chip in the header** — `waiting for you` / `turn ended` / `state unknown`, beside the existing live/idle/cold dot.
- **Amber strip** naming the tool and argument the agent is blocked on (`Bash: sw_vers`, `AskUserQuestion: which cohort?`). Pulses once on the transition, honours `prefers-reduced-motion`.
- **`<title>` marker and favicon dot**, so a background tab shows it without being read.
- **Opt-in system notifications** behind an "Enable alerts" button.
- **`.claude/hooks/agent_state.py`** — writes `projects/<id>/.agent-state.json` on `PermissionRequest`, `Notification`, `Stop` and `UserPromptSubmit`; `SessionEnd` clears it via `dash_stop.py`. Gitignored.
- **A hidden tab now polls** at 15s instead of skipping the fetch entirely — without it none of the above can update while you are away.

## Clearing

There is no "the prompt was answered" event. `PreToolUse` fires *before*
`PermissionRequest`, so the first thing that happens only because a human said yes is
`PostToolUse`. Rather than enumerate that, anything which is **not** one of the four
writing events clears — the agent only moves on once it has been unblocked, so this
covers future events too.

`Notification` is the exception, and the reason the rule is not "any event fired": it
arrives ~6s *into* an unanswered prompt. Its two stock messages are ignored entirely,
because both restate the state they are attached to — `idle_prompt` in particular fires
60s after every `Stop`, which otherwise turned each finished turn into a detail-less
"waiting for you", notification included, once per turn.

## Behaviour worth knowing

`PermissionRequest`, not `Notification`, supplies the detail: the notification payload is a fixed `"Claude needs your permission"` string with no tool in it, and it arrives ~6s later. `PermissionRequest` stays silent for auto-allowed calls, so it only fires when a human is really being asked.

Notifications stay quiet by default. `waiting` always notifies; `turn_ended` only when the tab has been hidden over 60s, since `Stop` fires at the end of every turn. Everything is keyed on `(state, since)` so a re-render never re-notifies.

A `waiting` claim expires: `unknown` after 30 minutes, or as soon as `runtime.json` has no record of the session that wrote it. Otherwise a killed session would leave the page saying "waiting for you" forever.

Detail text is agent-authored, so it goes through the same `inline_md`/`e()` escaping as the worklog.

## Limits

Snapshot mode has no poll — it reports the state it was written with and updates on reload. A closed tab gets nothing; that needs a service worker and a push service. Browsers throttle hidden-tab timers to ~1/min after ~5 minutes, so expect up to a minute of latency when away.

## Tests

450 pass, 22 new. Notification and pulse logic runs under node against a DOM stub (skipped if node is absent) rather than being asserted against the source; the page was also checked in headless Chrome. Design notes in `docs/live-dashboard-design.md`.

Replaces #351, which GitHub closed when the branch was renamed from `dileep-kishore/agent-notifications`.


tools/dashboard.py
```diff
@@ -902,6 +1009,67 @@ class on the overlay rather than given separate overlays — one close path. */
...
+# `#d-detail` is agent-authored text, so it is *not* interpolated here. The
+# server renders it through the same `inline_md` -> `e()` path as the worklog and
+# this copies the resulting node's HTML verbatim; the escaping decision stays in
+# one place, in Python, where it is tested. The notification body takes
+# `textContent` instead, since it is not markup at all.
+#
+# The `(state, since)` pair is the debounce key. Without it every 4s re-render is
+# a fresh transition: the strip would re-pulse and the OS notification would
+# re-fire for one permission prompt, which is how a notification gets muted.
+STATE_JS = """
```
briehl, line 1032: Yikes, that's hard to read. Should probably be part of a separate JS stack for the templates, or just hard coded into it.

tools/dashboard.py
```diff
@@ -914,10 +1082,24 @@ class on the overlay rather than given separate overlays — one close path. */
...
+# `scan()` at 6.8ms measured, so 15s hidden is ~1.6s of CPU per hour per tab,
+# less than a *visible* tab costs today.
+#
+# The single `timer` handle is not decoration. `tick` schedules the next tick,
+# and the visibilitychange listener calls `tick` directly, so every return to
+# the tab used to start a second concurrent chain that never ended — harmless
+# while hidden ticks were free, a compounding multiplier on real requests now.
 POLL_JS = """
 if(!location.pathname.endsWith('/'))location.replace(location.pathname+'/');
+var timer=0;
```
briehl, line 1101: Should probably be `let timer=0` but hard to tell with how intertwined it is.

briehl (approved):

PR #371, opened 2026-08-05
Add context manager to submit skill

briehl:
Adds a best-effort mirror step to `/submit`: after a project is archived to the
lakehouse, its files are also submitted to the BERIL-managed context service so
the knowledge layer can see the completed project.

The mirror is deliberately subordinate to the archive. The lakehouse copy — not
the context index — is the source of truth for "submitted", so a context service
that is down, or a user who isn't logged in, must never fail an otherwise-clean
upload. In those cases the tool reports the outcome and moves on.

## How it works

`tools/lakehouse_upload.py` gains `submit_to_context_service()`, called only on a
clean single-project archive. It shells out to `knowledge/scripts/ingest_context.py`
via `uv run --script` rather than importing it. That indirection is the main design
decision here: the upload tool runs under whatever interpreter the caller has
active — frequently one without the `openviking` SDK, e.g. the webapp venv —
whereas the ingest script's PEP-723 header pins its own environment. Shelling out
keeps the knowledge-layer dependencies out of the upload path entirely and lets the
mirror provision itself.

Note that this is a quick-and-dirty design choice and will be revamped in a later PR.
That PR will officially merge the knowledge layer into the beril codebase a little more
completely rather than keeping as a separate environment. It's just bigger than I 
want to submit all in one space.

`ingest_context.py` gains a `--json` verdict mode (requires `--project`) that gates
on three conditions, all required:

1. the BERIL webapp is available,
2. the user is logged in with a valid credential, and
3. the context service is reachable and accepts that credential.

(1) and (2) are proved together by an authenticated health call; (3) by the context
client's own reachability/auth diagnosis against the credential the importer will
actually use. Verdict mode prints a single line of `{"status", "reason"}` on stdout
and **always exits 0** — it is advisory, and it shares `ingest_project` with the
interactive path so the two modes can't drift.

## Contract changes

- `upload_project()` takes `mirror_to_context=False`. Off by default so batch and
  list callers don't trigger a re-ingest; the single-project CLI path turns it on,
  and `--no-context-mirror` opts back out. `--all` never mirrors.
- The upload tool's JSON payload gains an optional `context_submission`
  `{status, reason}` field. It does **not** affect the exit code — a skipped or
  failed mirror on a clean upload still exits 0. The existing 0/1/2 contract is
  unchanged, so `/submit` needs no new handling.
- Mirroring is skipped on a `warning` (partial) archive: the archive is incomplete,
  so there's nothing trustworthy to mirror yet. The caller treats that as a
  submission failure and retries, which re-runs the mirror on the next pass.

## Tests

Adds `tests/test_lakehouse_upload.py` — 61 tests covering the upload tool, which
previously had none. `mc` is stubbed with a fake that models the destination prefix
and records the call sequence, so the integrity behaviours are assertable without a
MinIO: SKIP_PATTERNS filtering (incl. `.submit.lock` and stale metadata), the
re-submission pre-clear and its abort-on-failure path, the manifest+1 partial-upload
check, metadata cleanup on both success and failure, the mirror gating above, and
the CLI's exit-code/JSON contract.

I mutation-tested the load-bearing invariants to confirm the tests bite — removing
the `+1` from `expected_remote_count`, disabling the pre-upload clear, and dropping
`.submit.lock` from `SKIP_PATTERNS` each fail exactly the tests written for them.
CI already runs `--cov=tools`, so this lands in the coverage report.


dileep-kishore (approved):
Looks good

PR #372, opened 2026-08-06
Record per-session agent cost and stamp per-stage ledger

dileep-kishore:
## Problem
BERIL had no way to know how much money was actually spent on a research project. The provenance system tracked who/what/when but nothing about cost, so `/runtime` and reviewers had no visibility into agent spend per stage of a project's lifecycle.

## Solution
The status line is the only component that ever sees a dollar figure (`cost.total_cost_usd` in its payload — no hook payload carries cost, and the transcript only has token counts), so it now persists that figure into each session's record in `runtime.json`. Separately, the runtime hook watches for `beril.yaml`'s `status:` field changing (the only code path that witnesses all six lifecycle transitions) and, when it does, stamps the *uncounted remainder* of session spend as a delta entry into a new `beril.yaml.agent_cost.stages[]` ledger, tracking `counted_usd` per session so a session spanning multiple stages is never double-counted.

The cost figure is explicitly labelled a floor: only Claude Code sessions are observed, work done by a human or another agent isn't counted, and an unwatched stage records `sessions_observed: 0` with `usd` omitted (never `0.00`) so it's distinguishable from a genuinely free stage.

Also extracted `block_span` (a shared YAML block-boundary scanner) out of `approve_cmd._drop_plan_approval` into `audit_cmd`, since both `plan_approval` deletion and `agent_cost` appending need to agree on the same block boundaries — and `audit_cmd` must stay off the CLI import chain so stage stamping still works under the bare `python3` fallback used on pods without a venv (no `tomllib`/3.11+ requirement).

`/runtime` and `docs/provenance-and-trust.md` were updated to describe the new `cost` and `agent_cost` fields and their floor semantics.

### Dashboard
The ledger is now visible on the project dashboard, since a number nobody reads is a number that may as well not be recorded.

Per-stage cost rides on the **stage rail**, which is already the list of stages in order — a stage's cost is per-rail-item data, not a second panel that would have to repeat the labels to say where each number belongs:

```
● exploring        $4.12
● plan written     $5.68
● running analysis   —          ← recorded, but nothing observed it
● report drafted
○ reviewed
○ approved
```

The **project total** sits in the header beside `last activity` / `elapsed`, which is the row you read before deciding whether to re-run an expensive notebook round.

Both carry the floor caveat in their `title` rather than in repeated captions, and both refuse to invent a zero: an unobserved stage renders an em dash, and a project with no observations at all (finished before this shipped, or worked by a human) gets no header total rather than a confident `$0.00`. A stage that legitimately repeats — `/synthesize` and `/berdl-review` both demote `reviewed → analysis` — sums rather than overwrites, so a first pass's spend isn't silently dropped.

`dashboard.py` parses `beril.yaml` by regex rather than importing `beril_cli`, matching the existing `plan_approval` parsing and the same bare-`python3`-in-a-pod constraint that makes `plan_digest` a deliberate copy. No cache work was needed: `fingerprint()` already stats every file, so a `beril.yaml` write invalidates the etag on its own.

### Testing
- `tests/test_cli_audit.py` covers the cost arithmetic (cross-stage attribution, unobserved-stage omission, snapshot-replace carrying cost forward, manifest-preserving append) and pins the import-chain constraint.
- `tests/test_statusline.py` adds a real end-to-end test running the actual hook and status line scripts through a full stage transition. Its docstring records what it does *not* cover: under `uv run pytest` the venv is on `PATH`, so the hook's bare-`python3` fallback finds a modern interpreter and the `tomllib` failure doesn't reproduce — verified by reverting the fix and watching it still pass. The deterministic import-chain guard is what catches that class; both are kept.
- `tests/test_dashboard.py` covers ledger parsing (including the repeated-stage sum and manifests with no block), the rail never rendering `$0.00` for an unobserved stage, and the header total omitting itself entirely when nothing was observed.

The dashboard rendering was additionally verified end-to-end against a `beril.yaml` produced by the real hook, rather than a hand-written fixture.


mikacashman (approved):
Looks good, its a nice breakdown.  Looks like it is project-centric so I wonder if exploitation use-cases also want to be able to see more details but I suspect that is an edge-case anyway and the status bar is enough for now (looks like the skill pulls from recent projects if you pull it up which is fine, could turn into an overload if several projects are accessible but future question).

PR #375, opened 2026-08-11
Add omp to supported coding agents

justaddcoffee:
## What

Adds [`omp`](https://www.npmjs.com/package/@oh-my-pi/pi-coding-agent) (oh-my-pi) as a launchable agent alongside `claude`, `codex` and `gemini`.

## Why the refactor first

The agent list was duplicated in three places (`cli.py` parser choices, `setup_cmd.py` detection, `doctor.py` detection). Adding a fourth agent meant editing three tuples that could silently drift. This moves it to a single `config.SUPPORTED_AGENTS`, and the new tests assert no module keeps a private copy.

## omp needs no special-casing

It discovers `.claude/skills` and reads `SKILL.md`, so `/berdl_start` and the other 15 BERIL skills resolve without changes. `start.py` already gated both the Vertex block and the model pin on `agent == "claude"`, so it needed no edit.

## Drive-by bug fix

`setup_cmd.py` passed `--model opus` **unconditionally** on launch:

```python
os.execvp(binary, [chosen, "--model", "opus", "/berdl_start"])
```

That hands an Anthropic-only flag to `codex` and `gemini` today, not just to `omp`. `start.py` already had the correct guard; `setup_cmd.py` now matches it. There's a regression test for both paths.

## Testing

- 15 new tests in `tests/test_cli_agents.py`
- Full suite: **173 passed**
- `build_parser()` split out of `main()` so the parser is testable directly (no behaviour change)

## Not in this PR

Wiring omp to the shared BERIL Vertex key. That is a larger piece of work: unlike Claude Code, omp takes a *static* `apiKey` string in `models.yml` and does not expand environment variables, so the short-lived Vertex OAuth token expires mid-session. It needs either a local refresh proxy or use of omp's own `auth-broker`. Worth doing separately. Users can point omp at their own key today.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ To use Claude Code you will need an API key. If you are Berkeley Lab staff, you 
 ### Prerequisites
 
 - A KBase account with BERDL access (see [Getting BERDL Access](#getting-berdl-access))
-- [Claude Code](https://claude.ai/claude-code) installed (or another supported agent e.g. codex)
+- [Claude Code](https://claude.ai/claude-code) installed (or another supported agent: `codex`, `gemini`, `omp`)
 - Python 3.11+
 - Git
 
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ cp .env.example .env
 # from .env automatically. If your token is ever exposed, delete it immediately
 # from the JupyterHub token management page and generate a new one.
 
-# 3. Open Claude Code in the repo
+# 3. Open your agent in the repo (claude, codex, gemini or omp)
 claude
 ```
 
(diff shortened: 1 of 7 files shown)


justaddcoffee:
Superseded by #376.

This branch was cut from a local `main` that was 334 commits behind upstream, so it conflicts, and its central change is obsolete: `claude_defaults()` already landed upstream and fixes the `--model opus` bug this PR was addressing, more thoroughly (it also handles `opus[1m]` and `--permission-mode`).

Rebasing onto current main would have meant discarding most of the diff, so #376 is a clean re-do against upstream HEAD. It keeps the omp support and the `SUPPORTED_AGENTS` consolidation, drops the now-redundant helper in favour of the existing `claude_defaults()`, and fixes a related bug still live on main: `claude_defaults(chosen, []) or ["--model", "opus"]` reinstates the flag for codex and gemini.

Issue #316, opened 2026-07-08
bootstrap_client.sh fails: pinned BERDataLakehouse repos (spark_connect_remote, berdl_remote) return 404


turbomam:
`uv run beril setup` (Step 4) and `bash scripts/bootstrap_client.sh` fail because the script installs two git dependencies that are no longer reachable:

```
git+https://github.com/BERDataLakehouse/spark_connect_remote.git
git+https://github.com/BERDataLakehouse/berdl_remote.git
```

Clone output:

```
remote: Repository not found.
fatal: repository 'https://github.com/BERDataLakehouse/spark_connect_remote.git/' not found
ERROR: bootstrap_client.sh failed (exit 1).
```

The whole `github.com/BERDataLakehouse` org now returns 404, not just these two repos, and I found no public relocation under `kbase`, `berdl`, `beril-doe`, or `kbaseincubator` (also nothing via GitHub code search). So the dependencies were deleted, renamed, or made private, most likely during the platform migration.

Effect: off-cluster BERDL client setup cannot complete, because `.venv-berdl` never builds. Anyone whose access does not already include the repos' new home is blocked at Step 4.

Suggested fixes:
- Point `bootstrap_client.sh` at the current locations of `spark_connect_remote` and `berdl_remote`, or publish them to PyPI / as wheels and pin those.
- If they are intentionally private now, document the required access in the off-cluster setup docs so the failure is expected rather than a dead end.

Related: #293 (off-cluster connection relies on undocumented workarounds), #315 (setup token prompt links to a page that shows no token).

turbomam:
Related symptom, plus a local-only workaround (not intended for commit).

The same dead pins break more than bootstrap. `scripts/run_sql.py` and `scripts/export_sql.py` declare `spark_connect_remote` and `berdl_remote` as PEP 723 inline dependencies pointing at the same `github.com/BERDataLakehouse` repos. So off-cluster, `uv run scripts/run_sql.py --berdl-proxy ...` (which the `berdl`, `berdl-query`, and `berdl-discover` skills call) fails with the identical 404, and the query skills are unusable until this is fixed.

Short-term local workaround, not intended for commit: if those two packages are already installed in a `.venv-berdl` (for example, reused from a prior working bootstrap), invoke the scripts with that venv's Python and bypass uv's resolution entirely:

```
.venv-berdl/bin/python scripts/run_sql.py --berdl-proxy --query "SELECT 1 AS ok"
```

[...]

PR #383, opened 2026-08-14
Never commit vendored BERDL packages to this public repo

turbomam:
A working checkout can accumulate `vendor/*.tgz` holding `berdl_remote` 0.1.0, `spark_connect_remote` 0.2.0 and `berdl_notebook_utils` 0.0.1, captured from a BERDL pod. Nothing in `.gitignore` covered them: no `vendor`, no `*.tgz`, and this repo has no `local/` convention either.

Those are third-party packages and this repository is public, so a single `git add -A` publishes them. That is a disclosure problem rather than a tidiness one, which is why this is a change rather than a note.

They exist locally because the two repos `scripts/bootstrap_client.sh` pins are unreachable (#316), so a captured copy is the only way off-cluster setup completes at all. Deleting them is not an option; ignoring them is.

Verified with `git check-ignore -v` rather than by reading the pattern:

```
$ git check-ignore -v vendor/ vendor/berdl-unreproducible-packages-2026-07-31.tgz
.gitignore:19:vendor/   vendor/
.gitignore:19:vendor/   vendor/berdl-unreproducible-packages-2026-07-31.tgz
```

`*.tgz` is included as well, since the same capture is easy to drop elsewhere in a checkout.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,13 @@
 docs/plans/
 docs/superpowers/
 
+# Vendored packages captured from a BERDL pod (see issue #316).
+# berdl_remote and spark_connect_remote are pinned from github.com/BERDataLakehouse,
+# which returns 404, so these local copies cannot be re-fetched. They are third-party
+# code and this repository is public: never commit them.
+vendor/
+*.tgz
+
 # Large data files
 *.npy
 *.pkl


briehl (approved):
Nice catch!

PR #388, opened 2026-08-19
Give omp a session directory a collector can name

justaddcoffee:
## Why

`beril start --agent omp` launches omp with no session flags, so omp writes its transcript to `~/.omp/agent/sessions/<encoded-cwd>/` under a name it picks — a timestamp and a UUID, knowable only after the fact. Because `beril start` chdirs to the repo root first, every project in a checkout also shares one such directory.

Claude Code gets a SessionStart hook that hands BERIL the exact `transcript_path` (`audit_cmd.py::_agent_signals_from_transcript`); omp fires no such hook, so a transcript reader has nothing to be pointed at. This is the BERIL half of adding omp support to evalome, whose `collect` takes `--transcript <path>`.

## What

Pass `--session-dir ~/.beril/omp-sessions/<name>-<hash>/` for omp and print it at launch, at both launch sites (`beril start` and `beril setup` step 11).

**In `$HOME`, not in the checkout.** BERDL compute nodes are ephemeral pods (PROJECT.md) while `$HOME` persists — `beril setup` says exactly that at step 10 — so a transcript in the working tree is the copy that does *not* survive a pod restart. It would also put full agent conversations inside a public repo's tree, one ignore-drift or `git add -f` from being committed, and inside the Docker build context. Keyed by hash of the resolved path, so two clones sharing a directory name do not share sessions.

**No mkdir.** omp creates the directory itself, recursively, on open (verified against the shipped binary). Creating it here would run between `print("Launching omp...")` and `execvp`, where a path occupied by a file or a read-only mount raises *after* the operator was told the agent was starting.

**Repo-root, not per-project** — `/berdl_start` scaffolds the project *during* the session, so at launch there is no project directory yet.

**Injected by BERIL rather than left to the operator.** The onboarding guard is `if not skip_onboard and not extra_args`, so an operator passing `--session-dir` by hand loses `/berdl_start`. Injecting after the guard means the default path gets both.

## Scope, stated honestly

This **routes around** the onboarding guard; it does not repair it. Any forwarded flag still drops `/berdl_start` — `--thinking high` does, and so does an explicitly passed `--session-dir`. That is pre-existing and now pinned by `test_any_forwarded_flag_still_costs_onboarding` so it is visible rather than implicit. Fixing it properly needs a per-agent "is this argv carrying a user message?" scanner, which is more machinery than this PR should carry.

Two more known gaps, both deliberate:

- `_has_flag` matches the flag appearing inside *message text*: `-p "--session-dir=x is what I mean"` suppresses the injected default. Same shared-substring weakness; same per-agent-parsing fix.
- `--session-dir` bypasses omp's `autoResume`. It is off by default, so this affects only operators who enabled it.

## Drive-by (same bug class — happy to drop it)

`claude_defaults` tested `"--model" not in extra_args`, which misses `--model=sonnet` and hands Claude the flag twice. `_has_flag` matches both spellings. Note the caveat: I could not verify how Claude Code resolves a repeated `--model`, so this is "stops passing a duplicate flag", not a demonstrated wrong-model bug.

## Verification

582 tests pass. Two mutations that previously survived the whole suite now fail:

- renaming the session-root constant — every assertion compared the constant to itself, so the old version passed while transcripts landed in a tracked path
- reverting the `--model=x` fix — only the helper was tested directly, with no end-to-end coverage

```
$ beril start --agent omp --dev
omp transcripts: ~/.beril/omp-sessions/BERIL-research-observatory-<hash> (this session is the newest *.jsonl)
argv: ['omp', '--session-dir', '<that path>', '/berdl_start']
```

Claude's argv is unchanged on the default path, and deliberately changed for `--model=x` / `--permission-mode=x`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -112,6 +112,10 @@ projects/*/.agent-state.json.tmp
 # ignoring it preserves the de-facto behaviour and stops a fresh clone showing an
 # untracked file after the SessionStart hook writes one.
 projects/*/runtime.json
+# omp session transcripts, when `beril start --agent omp` points omp at the checkout
+# (beril_cli/start.py::omp_defaults). Same genus as runtime.json: machine-local, and
+# the full conversation including whatever the agent happened to read.
+.omp-sessions/
 
 # UI application
 ui/data/cache.json
--- a/.gitignore
+++ b/.gitignore
@@ -112,10 +112,6 @@ projects/*/.agent-state.json.tmp
 # ignoring it preserves the de-facto behaviour and stops a fresh clone showing an
 # untracked file after the SessionStart hook writes one.
 projects/*/runtime.json
-# omp session transcripts, when `beril start --agent omp` points omp at the checkout
-# (beril_cli/start.py::omp_defaults). Same genus as runtime.json: machine-local, and
-# the full conversation including whatever the agent happened to read.
-.omp-sessions/
 
 # UI application
 ui/data/cache.json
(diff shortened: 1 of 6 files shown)


briehl (approved):
This looks great!

PR #389, opened 2026-08-20
Consolidate test suites and uv dependency groups

briehl:
Step 1 of the code consolidation, stacked on top of #371 (the context-service mirror). Targets `beril-submit-to-ov` rather than `main`, so the diff here is only the test and dependency work.

## Reuniting a package with its test suite

`knowledge/tests/` held the *only* tests covering `observatory_context/` — every file there imports `observatory_context.*` and none import anything from `knowledge/`. They ran under a separate CI job with `--cov=knowledge`, a directory with no importable library modules, so `observatory_context` was appearing in no coverage report anywhere.

- `knowledge/tests/` → `tests/knowledge/`, and the CLI tests → `tests/beril_cli/`.
- `run-knowledge-tests.yml` deleted; `run-cli-tests.yml` now covers both, with `--cov=observatory_context` added so that package is finally measured (81%).
- The `test` and `knowledge` dependency groups collapse: `pyyaml`/`boto3`/`openviking`/`rich` become hard `[project.dependencies]`, and test deps move to the default `dev` group so a plain `uv sync` is enough in CI.

**736 tests pass — exactly 625 (old `tests/`) + 111 (old `knowledge/tests/`), so nothing was lost or silently skipped.** No duplicate module basenames, and `uv sync` installs the `dev` group by default (verified with a clean-room resolve, not just locally).

## Why hard deps rather than a `beril[knowledge]` extra

This costs ~144 packages / 585M on install, so it's a deliberate call rather than an oversight:

- **The lean-install population is ~empty.** The only documented install is `pip install -e .` from a repo clone — there's no `uv tool install`/`pipx` path anywhere in the repo. Anyone installing has already cloned `observatory_context/` and `knowledge/`, and `/submit` needs the `openviking` SDK regardless.
- **An extra would be scaffolding with a known removal date.** Ingest is headed behind the BERIL webapp, behind an abstraction so the context manager is swappable. At that point the client stops importing `openviking` at all — the dependency *leaves* rather than gets re-partitioned. Adding an extra now means teaching docs and CI a boundary, then deleting it two steps later.

Worth revisiting only if a standalone `beril` distribution becomes a real target *before* the webapp migration lands.

## Follow-on: `--group knowledge` cleanup

Collapsing the groups makes `uv run --group knowledge` hard-error (`Group 'knowledge' is not defined`), so all 8 references are fixed. Two were user-facing error hints printed *when something is already failing* — `openviking_client.LOCAL_START_HINT` and the cancel script's restart panel — which would have sent someone debugging a down server into a second, unrelated error. One was prose in `docs/openviking.md` describing the group as something you'd otherwise need to activate, so it's reworded rather than flag-stripped.

## Review notes

- **The `uv.lock` diff is ~4700 lines** because collapsing the groups re-resolved everything. Nothing hand-edited; skip it.
- Most of the remaining diff is pure file moves (`git` shows them as renames with 0 changed lines).
- `--cov=knowledge` is kept alongside `--cov=observatory_context` for now. It measures exactly one file (`setup_remote_ov.py`, 43%) since the rest of `knowledge/` is PEP-723 scripts run as subprocesses; it can be dropped once those move server-side.


briehl:
Note that this is a chain with #371 - once that's merged, openviking and other knowledge stack things are basically required, so having different uv groups is less helpful.

briehl:
I'm not sure what the deal is with codecov in this case. With the test path merging and consolidation, there might be a tweak in how coverage percentage was calculated?

dileep-kishore (approved):

PR #396, opened 2026-08-20
Upgrade OpenViking pin 0.3.24 → 0.4.15

briehl:
## What

Pins every OpenViking declaration to `==0.4.15` (from `0.3.24` in the Dockerfile / unpinned elsewhere) and regenerates `uv.lock` (was resolving `0.3.14`). Server, CLI, and SDK move in lockstep per the release-note guidance.

Pin sites:
- `knowledge/openviking/Dockerfile` — server image
- `pyproject.toml` — top-level dependency
- the 5 PEP 723 inline headers in `knowledge/scripts/*.py` (run via `uv run --script` in isolated venvs, so they pin independently of `uv.lock`)

## Why 0.4.15 specifically — and never 0.4.14

0.4.14 declared an **unbounded `xxhash`** dependency. Installing `xxhash` 4.x makes the **local VectorDB backend (the one BERIL uses)** silently fail to persist newly generated vectors: the ingest task reports success and the source content is saved, but no vector lands, so later `find`/recall misses it **with no error surfaced**. 0.4.15 (upstream PR #3995) pins `xxhash` to fix this. The regenerated `uv.lock` resolves `xxhash 3.7.0`, confirming the fix.

## Why no client-code changes were needed

BERIL uses only the `viking://resources/…` namespace and the default `local` backend, so the two breaking changes since our last version — the User/Peer model (0.4.1) and strict-AND retrieval `tags` (0.4.11) — don't apply. SDK introspection against the installed 0.4.15 package confirms the surfaces BERIL depends on are unchanged (new params are additive with defaults):

- `SyncHTTPClient(url=, api_key=)` + `.initialize()` / `.get_status()`
- `find(query, target_uri, limit, filter, score_threshold, …)` and `add_resource(path, to, reason, wait, …)`
- all 18 client methods BERIL calls, the 8 `openviking_cli.exceptions` names, and `openviking.utils.search_filters.merge_time_filter`

## Verification

- `openviking-server doctor` → **Config PASS** (schema unchanged; only failure is the placeholder embedding key in `ov.conf.example`, expected).
- All **111** knowledge tests pass; every `observatory_context` OV module imports cleanly against 0.4.15.
- Live-checked against **beril-dev** (already serving **0.4.15**): `POST /api/v1/search/search` over `viking://resources/projects/` returns real project data in the 0.4.x `{status, result:{memories, resources, skills, total}}` envelope, with the additive empty `tags: []` field exactly as predicted.
- Docs/test fixtures bumped to 0.4.15 (`DEPLOYMENT.md` health sample, `test_diagnose.py`).

## Out of scope

The deployed **prod** SPIN server migration (still on 0.3.24) is a separate manual runbook — including the staged `0.3.24 → 0.4.5 (migrate+cleanup) → 0.4.15` path *if* the workspace holds legacy `viking://agent`/`viking://session` data (BERIL only ever wrote `viking://resources/`, so it likely does not, but the `ov ls viking://agent` gate decides). Not part of this PR.


briehl:
Despite the apparent size of changes, that's really just uv.lock getting an overhaul. The rest is just version pinning, and trivial. The build process has to occur outside (right now), and has already been done and deployed successfully, so I'll just go ahead and merge.

PR #403, opened 2026-08-25
Add OpenViking search page

briehl:
## What

A minimal **Search** page in the webapp that runs a semantic query against OpenViking: a text field, an optional max-results input, and result cards showing each hit's **summary (abstract)**. The goal is to get the end-to-end flow in place (page → JSON endpoint → OV → rendered summaries); iteration to follow.

This is also the **first properly committed React island** — establishing a reusable, committed esbuild build under `ui/components/search/`. (The chat island uses the same pattern but its build source was never committed; this can serve as the template to revive it.)

## How it works

**Backend**
- `ui/app/clients/openviking.py` — new `ov_search()` → `POST /api/v1/search/search` authed with the user's key (`X-API-Key`), reusing the existing `{status, result, error}` envelope parsing. 30s timeout (search embeds the query server-side, slower than the admin calls).
- `ui/app/routes/search.py` (new) — `GET /search` (SSR shell) + `GET /api/search` (JSON). Queries **as the logged-in user** (`get_ov_credential` + `decrypt_secret`). Limit clamped to 1–50. Targets `viking://resources/projects/`.
- `ui/app/main.py` — registers `ROUTER_SEARCH`.

**Frontend**
- `ui/components/search/` — React 19 + esbuild 0.25 (matching chat's versions), a `build` script, `src/index.jsx` (`window.BERILSearch.mount`), `src/SearchApp.jsx`, and a README documenting the build. `node_modules` gitignored; source + built bundle committed.
- `ui/app/templates/search.html` (new) mounts the island via dynamic import, mirroring `chat/session.html`. `base.html` gets a top-level "Search" nav link.

## Graceful failure (never a traceback)

The JSON endpoint returns `{error, message}` for every expected failure, rendered as an inline notice: empty query (400), **no OV credential yet** (409, prompts the user to provision on their profile), decrypt failure (500), rejected key (502 `auth_failed`), other OV/transport errors (502).

## Query mechanics (verified against openviking 0.4.15)

The default find/list path makes **exactly one external call — embedding the query string** (Vertex AI `gemini-embedding-001`). **No generative/VLM call**: abstracts are read back from ingest-time storage. The LLM intent-analysis branch only runs when a `session_id` is passed, which this endpoint never does. (Implication for later: no search-as-you-type without debouncing, since each query is a paid embedding call.)

## Verification

- App builds; `/search` and `/api/search` register; bundle installs `window.BERILSearch`.
- All `api_search` branches unit-tested (empty / no-credential / happy / auth-failed).
- **Not yet done:** full browser e2e against beril-dev (needs a logged-in session + `BERIL_OV_URL` pointed at dev). Safe to review/iterate given the graceful-fallback coverage.

## Known follow-ups for iteration
- Decide **per-user key vs service account**: today a user must have provisioned an OV account (`POST /api/ov/user`) or search returns the "set one up" notice. A shared service key would make it work for any logged-in user.
- Live e2e verification of the `X-API-Key`/search-endpoint auth against beril-dev.


.gitignore
```diff
@@ -117,6 +117,10 @@ projects/*/runtime.json
 ui/data/cache.json
 ui/data/indexdir/
 ui/user_projects/
+# React island build deps — source under ui/components/<name>/src/ and the built
+# bundle under ui/app/static/<name>/ ARE committed; only node_modules is ignored.
+ui/components/*/node_modules/
+ui/static/search/*
```
dileep-kishore, line 123: I think this should be `ui/app/static/search/*` (same as dockerignore)?

dileep-kishore (changes requested):
I tried a few searches and noticed that many of the results are `.abstract.md` and `.overview.md` files, or chunks extracted from larger files. The project grouping and links are helpful, but viewing the matched chunk or loading the full file will be important for making the results useful beyond contextual retrieval. Since that is already planned for the next PR, I think the current version is a reasonable first step.

Do we want unit tests for any of the new UI components? Codecov is flagging coverage, although it appears to be the Python search route and client

PR #404, opened 2026-08-25
Bump GitHub Actions pins to latest majors

briehl:
Every action in our workflows was pinned to a version running on the deprecated node20 runtime, which GitHub Actions has been warning about. This bumps all 16 pins to the current latest majors and adds a Dependabot config so they don't drift again.

## Pin changes

| Action | Before | After |
|---|---|---|
| actions/checkout | v4 (×3) / v7 | v7 |
| actions/setup-python | v5 | v7 |
| astral-sh/setup-uv | v4 | v10.0.1 |
| codecov/codecov-action | v5 | v7 |
| docker/setup-buildx-action | v3 | v4 |
| docker/login-action | v3 | v4 |
| docker/metadata-action | v5 | v6 |
| docker/build-push-action | v5 | v7 |

This also resolves an inconsistency where `actions/checkout` was pinned to v4 in three workflows and v7 in a fourth.

`setup-uv` is pinned to an exact release rather than a floating major: the project publishes releases through v10.0.1 but stopped maintaining floating major tags after v7, so `@v10` does not resolve. Dependabot covers the resulting patch drift.

`deploy-spin-dev.yml` and `deploy-spin-prod.yml` are untouched — they use only `run:` steps and no actions.

## Breaking changes reviewed

Most of these majors are node20→node24 runtime bumps, which are safe on GitHub-hosted `ubuntu-latest` (they would only break self-hosted runners older than v2.327.1, and we have none). Three warranted a closer look:

- **setup-uv v4→v10.0.1** — v10 disables caching by default on `pull_request_target`, `workflow_run`, and `release` events as a cache-poisoning fix. Our uses trigger on `pull_request` and `push`, and we never set `enable-cache`, so nothing to migrate.
- **codecov v5→v7** — the `file`→`files` / `plugin`→`plugins` renames landed in v5, which we were already on. Our existing inputs are still current.
- **build-push-action v5→v7** — drops two deprecated environment variables and legacy export-build support; we use neither. v6 added build summaries, so image builds will now attach one to the job (cosmetic).

Also worth noting: `actions/checkout@v7` blocks checking out fork PR code under `pull_request_target` and `workflow_run`. `deploy-spin-prod.yml` triggers on `workflow_run` but never checks out code, so it is unaffected.

## Dependabot

Adds `.github/dependabot.yml` with a weekly `github-actions` ecosystem check, so future bumps (including minors like checkout v7.0.0→v7.0.1) arrive as PRs.

## Testing

The pin changes are static and CI is the real test. Opening this PR exercises 12 of the 16 pins via `run-cli-tests` and `run-webapp-tests`, which trigger on `pull_request`. The remaining four are in `build-image.yml`, which needs a push to `main` or a manual dispatch — worth a `workflow_dispatch` run before merge if you want full coverage.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/workflows/build-data-cache.yml
+++ b/.github/workflows/build-data-cache.yml
@@ -22,7 +22,7 @@ jobs:
           python-version: '3.12'
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v10
+        uses: astral-sh/setup-uv@v4
 
       - name: Install dependencies
         run: |
--- a/.github/workflows/build-data-cache.yml
+++ b/.github/workflows/build-data-cache.yml
@@ -12,17 +12,17 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v7
         with:
           fetch-depth: 0
 
       - name: Set up Python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v7
         with:
           python-version: '3.12'
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v4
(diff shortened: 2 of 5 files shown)


dileep-kishore (approved):
Looks good!

PR #411, opened 2026-08-28
Pin openviking-sdk in PEP723 knowledge scripts

### LINKED ISSUE #402: PEP723 knowledge scripts pin openviking but not openviking-sdk, so add_resource loses its reason argument
`knowledge/scripts/ingest_context.py` fails with `TypeError: add_resource() got an unexpected keyword argument 'reason'`, while the identical call succeeds from the project venv.

Found by the `was-langfuse` Claude session on 2026-08-24, running it against a live OpenViking server. Root cause below verified by direct measurement on 2026-08-24, not inferred.

## Cause

`openviking.SyncHTTPClient` is a compatibility shim (`openviking_cli/client/_http_compat.py`) that delegates to `openviking_sdk`. The shim file is byte-identical in both environments, sha256 `f24924dfac4db637`, so the differing code is in the SDK underneath it.

`openviking` does not put a ceiling on that SDK. From the 0.4.15 wheel metadata:

    Requires-Dist: openviking-sdk>=0.1.1

So:

| environment | openviking | openviking-sdk | `add_resource` params |
|---|---|---|---|
| project venv (via `uv.lock`) | 0.4.15 | **0.1.8** | 21, includes `reason` |
| PEP723 script env | 0.4.15 | **0.1.9** | 7, no `reason` |

`uv.lock` pins `openviking-sdk` to 0.1.8. A PEP723 script has no lock, so `uv run --script` resolves the newest SDK, and 0.1.9 dropped `reason`. Pinning `openviking==0.4.15` in the script block, which these already do, is necessary but not sufficient.

The scripts then `sys.path.insert` the repo and call `observatory_context/ingest.py:321`, which passes `reason=`. Repo code, floating transitive dependency.

## Affected

All five scripts carry `openviking==0.4.15` with no SDK constraint, and all import `observatory_context`:

- `knowledge/scripts/ingest_context.py`
- `knowledge/scripts/check_openviking_health.py`
- `knowledge/scripts/knowledge_query.py`
- `knowledge/scripts/smoke_ingest_openviking.py`
- `knowledge/scripts/setup_remote_ov.py`

## Done when

Each script block also pins `openviking-sdk` to the version `uv.lock` resolves, and `ingest_context.py` completes an ingest against a running server.

---

Body corrected 2026-08-24. The first version of this issue claimed the scripts declared `openviking` unpinned. That was wrong: it described a stale branch, not `main`, which pinned `openviking==0.4.15` in commit `2928c46f`. `was-langfuse` caught it. The pin is real and the bug is real; the mechanism is the unpinned transitive SDK, described above.

briehl:
Fixes #402

## Problem

The five `knowledge/scripts/*.py` PEP723 scripts pinned `openviking==0.4.15` but left the transitive `openviking-sdk` floating — `openviking` declares only `openviking-sdk>=0.1.1`.

The project venv gets 0.1.8 via `uv.lock`. A PEP723 script has no lock, so `uv run --script` resolved the newest SDK, and 0.1.9 dropped the `reason` parameter from `add_resource()`. `observatory_context/ingest.py:322` passes `reason=`, so it raised `TypeError: add_resource() got an unexpected keyword argument 'reason'` under the script envs while succeeding from the project venv.

## Fix

Add `"openviking-sdk==0.1.8"` to each script's dependency block, matching what `uv.lock` resolves:

- `knowledge/scripts/ingest_context.py`
- `knowledge/scripts/check_openviking_health.py`
- `knowledge/scripts/knowledge_query.py`
- `knowledge/scripts/smoke_ingest_openviking.py`
- `knowledge/scripts/setup_remote_ov.py`

## Verification

Reproduced the issue's table by direct measurement:

| env | openviking-sdk | `add_resource` params | `reason` |
|---|---|---|---|
| unpinned (before) | 0.1.9 | 7 | absent |
| pinned (after) | 0.1.8 | 21 | present |

All five scripts now resolve `openviking-sdk 0.1.8` in isolated script envs and import / `--help` cleanly.

## Not covered

The issue's second "done when" condition — `ingest_context.py` completing an ingest against a running server — is **not** verified here; this was developed off-hub with no OpenViking server reachable. The failing call path and the restored signature are both confirmed, but the end-to-end run should be checked before merge.

Note: the pin is a hard `==` to match the lock, so bumping the SDK now means updating the lock and these five blocks together.

### DIFF
--- a/knowledge/scripts/check_openviking_health.py
+++ b/knowledge/scripts/check_openviking_health.py
@@ -3,6 +3,7 @@
 # requires-python = ">=3.11"
 # dependencies = [
 #     "openviking==0.4.15",
+#     "openviking-sdk==0.1.8",
 #     "pyyaml",
 #     "rich",
 # ]
--- a/knowledge/scripts/ingest_context.py
+++ b/knowledge/scripts/ingest_context.py
@@ -3,6 +3,7 @@
 # requires-python = ">=3.11"
 # dependencies = [
 #     "openviking==0.4.15",
+#     "openviking-sdk==0.1.8",
 #     "httpx",
 #     "pyyaml",
 #     "rich",
--- a/knowledge/scripts/knowledge_query.py
+++ b/knowledge/scripts/knowledge_query.py
@@ -3,6 +3,7 @@
 # requires-python = ">=3.11"
 # dependencies = [
 #     "openviking==0.4.15",
+#     "openviking-sdk==0.1.8",
 #     "pyyaml",
 #     "boto3",
 # ]
--- a/knowledge/scripts/setup_remote_ov.py
+++ b/knowledge/scripts/setup_remote_ov.py
@@ -4,6 +4,7 @@
 # dependencies = [
 #     "httpx",
 #     "openviking==0.4.15",
+#     "openviking-sdk==0.1.8",
 # ]
 # ///
 """One-time setup for the remote OpenViking knowledge server.
--- a/knowledge/scripts/smoke_ingest_openviking.py
+++ b/knowledge/scripts/smoke_ingest_openviking.py
@@ -3,6 +3,7 @@
 # requires-python = ">=3.11"
 # dependencies = [
 #     "openviking==0.4.15",
+#     "openviking-sdk==0.1.8",
 #     "pyyaml",
 #     "rich",
 # ]


turbomam (approved):
I ran `knowledge/scripts/smoke_ingest_openviking.py` against a local OpenViking server twice: once with the `openviking-sdk==0.1.8` line present in the dependency block, and once with that line deleted.

Without the new line:

> `TypeError: SyncHTTPClient.add_resource() got an unexpected keyword argument 'reason'` at `observatory_context/ingest.py:321`

With the line back in, the ingest ran and I could query the results.

I'm new to PEP723 blocks. Is that adequate verification?

PR #419, opened 2026-09-02
First step toward integrated context manager (1/4?)

briehl:
This starts the push toward putting the context manager (OpenViking in this case, possibly others) behind an abstraction layer in the BERIL webapp API. I expect it'll be 4 PRs to bring in the needed functionality, plus one more to remove older code and scripts

1. (this one) initial scaffold and design, include search query and file listing
2. fill out that scaffold, include first ingest step, direct to the context manager, rewrite skills and scripts
3. expand on the ingest so files land somewhere they can be quickly retrieved
4. populate a file browser (mostly already exists from previous work) based on just the files in the system, not in the repo

The previous design separated concerns a little too much - the BERIL webapp was used for handling authentication and health checks, while OpenViking was installed and exposed separately. This caused confusion for Claude and the skills it uses - it checks to see if OV is alive, and if not, it tries to install it locally. That's... not ideal. IMO the user / agent shouldn't have to know or care whether there's something in the background doing any context searches or not. They should just do their work, submit and share when complete, and reap the results of everyone else doing the same. 

The goal with this design is to integrate the context manager into BERIL itself. It will include new endpoints for file ingest and searching that return some regularized results that mostly match the same results from OpenViking, though with a slightly different structure.

This first PR focuses on the initial scaffold and redesign.
Includes the following routes:
`/api/context/find` - runs the `find` command over the existing context manager rag db, which does a quick semantic search without session context. Including an ongoing session isn't really a goal yet, but once we get this up and stable we can introduce it.
`/api/context/ls` - runs the `ls` command, which really just shows the list of projects in the context manager right now with the timestamp of when it was last modified. Internally, this will be used to align changes with the ingest. This isn't as useful as the `tree` command, which will get included later.


dileep-kishore:
Agree with the goal and happy to move this in. Three things before merge. Rebase on main; this PR's only conflict is the Langfuse router in `ui/app/main.py`. The `pyproject.toml` rewrite (re-sort, pillow removal, pytest and ruff to dev) is unrelated churn in a foundational PR; I'd split it out. And at this PR's own head, `get_context_manager` dereferences `cred.encrypted_key` when no credential row exists, so first use 500s until #421 lands. Either fold that in or note this isn't deployable alone.

One design ask for the stack: make `ContextManager` in `base.py` declare the methods the routes actually call (its `list_files` signature already differs from the implementation, and routes import a dozen OpenViking-specific names directly), and decide one read-scoping policy. `/find` forwards `root_path` and `filter` verbatim while every later route refuses a raw URI.

PR #421, opened 2026-09-03
automatically resolve openviking credentials for users (2/4?)

briehl:
This one's a little under the hood, but another step toward OpenViking abstraction.

The BERIL webapp now automatically creates OpenViking users and stores / retrieves / updates their credentials as needed to support the /api/context paths. Those credentials don't need to be, and shouldn't be, exposed outside of the webapp. So now they aren't!

### DIFF
--- a/ui/app/context_manager/base.py
+++ b/ui/app/context_manager/base.py
@@ -3,8 +3,6 @@ from pathlib import Path
 
 from pydantic import BaseModel
 
-from app.db.models import BerilUser
-
 
 class FileMetadata(BaseModel):
     created: datetime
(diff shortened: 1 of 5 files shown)


dileep-kishore:
Two concurrent first-use requests can leave a revoked key stored: A registers, B gets 409 and regenerates (invalidating A's key), then A's upsert commits last. After that, reads 502 and ingest reports per-file failures, with no automatic repair; `POST /api/ov/user/regenerate` still works but nothing points a user there. Suggest serialising provisioning and rotation per user, and on an OV auth error with a stored key, regenerating once and retrying. Also `CredentialEncryptionError` from `decrypt_secret` is unhandled here and surfaces as a 500.

PR #423, opened 2026-09-08
ContextManager insert command

briehl:
Adds the insert command for the context manager. This deprecates the OpenViking scripts.

It also has a few key changes.
1. It lays groundwork for storing uploaded files, in whole, in the webapp file system. This will have some devops implications - like making sure things are in a stable storage location that gets backed up.
2. In the OpenViking paths, to maintain ownership, etc., this adds the user orcid to the file path. Now, for each file in OV, the resource URI becomes:  
`viking://resources/users/<orcid>/<project id>/path/to/file.xyz`


dileep-kishore:
This is where I'd like to settle gating. The route accepts any valid archive from any token holder and creates the `UserProject` row on the fly. `beril.yaml` is already staged, so the route can require `status: complete` and an `approval` block, match `approval.by` to the caller's ORCID, and check `sha256(REPORT.md)` and `sha256(REVIEW.md)` against `report_hash` and `review_hash` (stored as `sha256:<hex>`). That's an attribution and integrity check rather than proof of review, but it closes the open door. `force` should go through the existing `require_admin`. Follow-up for `/submit`: the approval block hashes notebooks but not `memories/*.md`, so memory files aren't yet bound to the approved report.

Archive handling: `_extract_archive` extracts every member before any size check, and the archive bytes are read with no cap. Please check `ZipInfo.file_size` and the aggregate expanded size before extracting, extract only manifest members, and cap the upload. `nginx.conf` sets no `client_max_body_size`, so anything over the 1 MB default gets a 413 at the proxy; the proxy, archive, and per-file limits should be set together.

The new `users/<orcid>/<slug>` URIs have no mapping in `fallback.py` or `berdl_fallback.py`, so the offline tiers return nothing for anything this route ingests.

Minor: the retry log call in `clients/openviking.py` passes a second f-string as a format argument.

PR #425, opened 2026-09-09
Route the /submit context mirror through BERIL's ingest API

briehl:
## Summary

The best-effort context mirror that runs after a successful lakehouse archive talked to OpenViking directly, so a user needed an OV key on their machine to submit a project. This routes it through BERIL's `/api/context/ingest_files` endpoint instead — the CLI now holds only a BERIL personal access token, and the backend's address, credential, and task vocabulary stay server-side.

Builds on the archive+manifest endpoint in 5cb50df2. A multipart upload loses relative paths (`memories/pitfalls.md` collapses to `pitfalls.md`), so the mirror zips the staged tree and sends a manifest naming which members to ingest.

## Changes

**New — `observatory_context/beril_ingest.py`**
- `build_archive()` — zips the staged tree, POSIX-relative paths preserved
- `submit_ingest()` — POSTs archive + manifest + project with the `beril login` PAT
- `poll_batch()` — polls `/api/context/ingest_status/{batch_id}` every 3s until terminal (10 min cap), retrying through transport blips since the batch is queued server-side regardless
- `ingest_project_files()` — the whole flow; returns an `IngestOutcome` separating success / per-file failures / timeout

**`knowledge/scripts/ingest_context.py`** — `--json` verdict mode only. Preflight drops from three gates to two (logged in, BERIL reachable); the backend-credential diagnosis is gone. Still calls `stage_project()`, so the generated `PROJECT_METADATA.md` and `CLAIMS_CONTEXT.md` ship as before.

**`.claude/skills/submit/SKILL.md`** — Phase 3b describes the archive-and-poll flow and its verdict mapping. No backend named anywhere in the skill.

## Scope and consequences

- **Interactive modes are unconverted.** `--all`/`--changed`/`--project`/`--docs` still drive the SDK directly, which is why the PEP-723 header keeps its `openviking` pin. The module docstring now says so instead of claiming the whole script targets OpenViking. Note this means `/submit` still installs an SDK it no longer uses — splitting the verdict mode into its own httpx-only script would fix that, deferred for now.
- **A timed-out poll reports `skipped`, not `failed`.** The files are queued server-side and may still land; the lakehouse archive remains the source of truth for "submitted".
- **Relations and the `knowledge/state/` manifest are skipped** — SDK-level operations with no route equivalent. A mirrored project's manifest entry stays whatever the last interactive run recorded, so a later `--changed` will re-ingest it. Re-run an interactive mode to reconcile.

## Testing

`tests/knowledge/test_beril_ingest.py` — 12 cases over `httpx.MockTransport`: nested-path preservation, manifest/member agreement, files outside the staged root, multipart shape, completion, failure-with-reasons, timeout, blip retry, and the missing-`batch_id` case. No BERIL instance or backend needed.

Full suite: 766 passed.

Writing these caught two bugs in the first draft — a caller-supplied client bypassed the auth header, and `ingest_project_files` swallowed `client` into `**poll_kwargs` so the submit call never received it. Auth is now attached per-request.

### DIFF
--- a/.claude/skills/submit/SKILL.md
+++ b/.claude/skills/submit/SKILL.md
@@ -203,7 +203,7 @@ Phase 2c is split into a **pre-write validation pass** (no state mutation) follo
 ##### 2c-pre. Validate inputs and stage memory extractions (no writes yet)
 
 - **Compute notebook hashes (v5)**: invoke `python {repo_root}/tools/notebook_hash.py compute-hashes {project_path}` to get a JSON dict of `{relpath: "sha256:<hex>"}`. Result is sorted by path, ready for stable YAML output. Empty dict for projects without `notebooks/`. Values are already `sha256:`-prefixed and can be written to YAML verbatim. (Read-only; safe to do here.)
-- **Stage memory extractions** from `REPORT.md`. This is how reviewed-and-approved discoveries/performance claims become OV-ingestible memories without contaminating the layer with unvetted synthesizer output. Validate and capture in memory only — no files are written or deleted in this pre-pass.
+- **Stage memory extractions** from `REPORT.md`. This is how reviewed-and-approved discoveries/performance claims become ingestible memories without contaminating the context layer with unvetted synthesizer output. Validate and capture in memory only — no files are written or deleted in this pre-pass.
 
   For each section name in `["Discoveries", "Performance Notes"]` (mapping to `discoveries.md` and `performance.md` respectively):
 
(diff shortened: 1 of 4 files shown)


dileep-kishore:
`poll_batch` catches `BerilIngestError`, but `http.get` raises `httpx.HTTPError` directly, so a dropped connection aborts the mirror on the first blip. The "transport blip" test uses a 503 response, which takes the other branch. Reproduced with `httpx.ConnectError`.

PR #426, opened 2026-09-10
Context Manager ingest should skip unchanged files

briehl:
When submitting a project, the context manager route scans all files to see if anything is unchanged, and only resubmits files to the context manager (OpenViking) if those files are changed.

This also involves tracking SHA-256 hashes for each ingested file, which is on the path to keeping local files stored for retrieval later.


dileep-kishore:
Skip-on-unchanged reads `completed_file_hashes`, which only sees rows whose latest record is `completed`. If the client stops polling before completion and OpenViking expires the task (24h completed, 7d failed), `get_task` returns `None`, `task_statuses` maps that to `unknown`, and the status route discards every `unknown`. Those files then re-ingest on every submit until some later batch is observed to complete. Distinguishing "backend unreachable" from "task gone" stops the indefinite polling, but doesn't restore the skip; for that, reconcile against OpenViking (`stat` on the target URI) before the skip check.

Also worth a stated position: re-submission only adds. A memory file withdrawn at re-approval stays searchable and still counts as "has memory" for #443.

PR #427, opened 2026-09-11
Extend /api/context/find toward query parity

briehl:
The route dropped most of what `knowledge_query.py find` can express: metadata filters, time bounds, a node limit, and full-document reads. Any skill needing those had to reach past BERIL to the backend directly, which is what keeps the backend's name in `.claude`.

ContextQuery gains filter, since/until, time_field, node_limit and read_content. QueryResult gains match_reason and content; results gain the backend's own total, which can exceed the rows returned under a limit. `text` still carries the abstract — content is the document, not a summary, so they are separate fields.

Limits are bounded rather than forwarded: a query asking the backend to walk its whole graph is a 422 before the backend is touched. Options are built by omission, since sending None would override a backend default the caller never chose — notably read_content=False, which is meaningful.

Query failures now surface as 502 instead of 500, via a QUERY_FAILURES tuple defined alongside the existing INGEST_FAILURES. It lives in the context_manager module deliberately: a route catching the SDK's own exception types would put the store's name back in the layer that exists to hide it. The client is also closed when find raises, which it was not before.

First of several route extensions; the CLI and skill rewrites depend on this surface existing.

### DIFF
--- a/ui/app/clients/openviking.py
+++ b/ui/app/clients/openviking.py
@@ -88,12 +88,49 @@ class OpenVikingClient:
     async def close(self):
         await self._client.close()
 
-    async def find(self, query: str, target_uri: str|None = None, limit: int = 10, score_threshold: float|None=None) -> dict:
-        options = None
-        if score_threshold is not None:
-            options = {"score_threshold": score_threshold}
-        result = await self._client.find(query, limit=limit, target_uri=target_uri, options=options)
-        return result
+    async def find(
+        self,
+        query: str,
+        target_uri: str | None = None,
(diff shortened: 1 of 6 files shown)


dileep-kishore:
`read_content=true` forwards a field the pinned OpenViking 0.4.15 `FindRequest` forbids (`extra="forbid"`, no such field), so it always returns 502. The test supplies a payload the server never produces. Either read content in the adapter after the find, or drop the option until then. And `root_path` and `filter` are forwarded verbatim while `/ls` and `/grep` refuse raw URIs; scope them the same way.

Issue #407, opened 2026-08-28
Access Request: BERDataLakehouse/spark_connect_remote


jinyun1tang:
I am trying to install BERDL client, but encountered 
 fatal: repository 'https://github.com/BERDataLakehouse/spark_connect_remote.git/' not found.
Can you get me access to the git repo indicated above? My GitHub name is: jinyun1tang

cmungall:
@jinyun1tang is trying to install the client on his local machine. I think this is not our preferred route, so I am working with him to run what he needs on BERIL.

cmungall:
**The org was renamed, not deleted: `BERDataLakehouse` → `KBaseDataLakehouse`.** This request should target the new org, or it will be actioned against one that 404s.

`KBaseDataLakehouse/berdl_docs` is public (HEAD `f8b7e33`, 2026-08-23) and names the current install line:

```
services/spark_connect_remote.md:24
pip install "git+https://github.com/KBaseDataLakehouse/spark_connect_remote.git"
```

Its README (lines 270–271) links both `KBaseDataLakehouse/spark_connect_remote` and `.../berdl_remote`.

`gh api orgs/KBaseDataLakehouse` resolves and reports `public_repos: 2`. Both packages return "Could not resolve to a Repository" to a token with `repo` scope. GitHub returns that identically for private and absent, so it does not prove they exist — but berdl_docs pointing at them last week is good evidence they do, and are private.

### The ask

Read access on `KBaseDataLakehouse/spark_connect_remote` and `KBaseDataLakehouse/berdl_remote` for: @jinyun1tang, @turbomam, @cmungall (at least — anyone doing off-cluster setup).

**Better, if it's cheap:** publish the two as wheels and pin those in `scripts/bootstrap_client.sh`. Setup then stops depending on anyone's org membership, which is what makes this recur. Exact versions are known — captured off a working pod on 2026-07-31 by @turbomam (#316):

| package | version |
|---|---|
| `spark_connect_remote` | 0.2.0 |
| `berdl_remote` | 0.1.0 |
| `berdl_notebook_utils` | 0.0.1 |

### Blast radius

[...]

PR #429, opened 2026-09-11
fix(scripts): repoint BERDL client deps at renamed KBaseDataLakehouse org

cmungall:
## Problem

The `BERDataLakehouse` GitHub org no longer resolves, so every pin to it is **unresolvable, not merely stale**. `bash scripts/bootstrap_client.sh` fails before `.venv-berdl` is built. Because `run_sql.py`, `export_sql.py` and `berdl_inventory.py` carry the same pins in PEP 723 headers, they re-resolve and 404 *at invocation* — which takes `/berdl`, `/berdl-query` and `/berdl-discover` down with them, a long way from the file that names the org.

## Root cause

**The org was renamed, not deleted: `BERDataLakehouse` → `KBaseDataLakehouse`.**

Source is the org's own public docs repo, `KBaseDataLakehouse/berdl_docs` (HEAD `f8b7e33`, 2026-08-23):

```
services/spark_connect_remote.md:24
pip install "git+https://github.com/KBaseDataLakehouse/spark_connect_remote.git"
```

Its README (lines 270–271) links both `KBaseDataLakehouse/spark_connect_remote` and `.../berdl_remote`.

This was the open question in #316, which tracked the breakage for seven weeks and concluded *"the organisation is gone… deleted, renamed, or made private"* after searching `kbase`, `berdl`, `beril-doe` and `kbaseincubator` — but not `KBaseDataLakehouse`.

## What this changes

All 11 references across 5 files:

| File | Refs |
|---|---|
| `scripts/bootstrap_client.sh` | 2 |
| `scripts/run_sql.py` (PEP 723) | 2 |
| `scripts/export_sql.py` (PEP 723) | 2 |
| `scripts/berdl_inventory.py` | 4 |
| `.claude/skills/berdl_start/SKILL.md` | 1 |

## What this does NOT do

**It does not unblock setup.** Both repos are private within the new org, so the installs still fail without read access. Rather than leave a correct-looking URL that fails for an unexplained reason, `bootstrap_client.sh` gets an inline note:

```bash
# NOTE: spark_connect_remote and berdl_remote are PRIVATE repos in the
# KBaseDataLakehouse org (renamed from BERDataLakehouse). Read access is required
# or the two installs below fail with "remote: Repository not found."
# Access request: .../issues/407
```

The access request is #407. Longer term, publishing the two as wheels would remove the org-membership dependency entirely — @turbomam captured the versions the platform ships from a working pod: `spark_connect_remote` 0.2.0, `berdl_remote` 0.1.0.

## Verification

- `bash -n scripts/bootstrap_client.sh` passes
- 11 `KBaseDataLakehouse` / 0 `BERDataLakehouse` remaining under `scripts/` and `.claude/`
- Cannot end-to-end test the install without org access — that is the point of this PR's caveat

Refs #407. Supersedes #316 (closed as duplicate). Docs companion: `docs/berdl-offcluster-access-prereq`.

### DIFF
--- a/.claude/skills/berdl_start/SKILL.md
+++ b/.claude/skills/berdl_start/SKILL.md
@@ -136,7 +136,7 @@ python scripts/berdl_inventory.py --refresh
 This is plain `python` in **both** environments. It auto-detects on-cluster vs off-cluster and picks the right discovery path.
 
 - **On-cluster (JupyterHub):** the JH kernel already has every import. Just run the command above.
-- **Off-cluster (local machine):** activate `.venv-berdl` first (`source .venv-berdl/bin/activate`), then run the command above. If the venv isn't bootstrapped yet, run `bash scripts/bootstrap_client.sh`. Ad-hoc alternative without venv: `uv run --with pyspark --with "spark_connect_remote @ git+https://github.com/BERDataLakehouse/spark_connect_remote.git" --with "berdl_remote @ git+https://github.com/BERDataLakehouse/berdl_remote.git" scripts/berdl_inventory.py`.
+- **Off-cluster (local machine):** activate `.venv-berdl` first (`source .venv-berdl/bin/activate`), then run the command above. If the venv isn't bootstrapped yet, run `bash scripts/bootstrap_client.sh`. Ad-hoc alternative without venv: `uv run --with pyspark --with "spark_connect_remote @ git+https://github.com/KBaseDataLakehouse/spark_connect_remote.git" --with "berdl_remote @ git+https://github.com/KBaseDataLakehouse/berdl_remote.git" scripts/berdl_inventory.py`.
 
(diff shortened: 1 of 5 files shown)


briehl (approved):

dileep-kishore:
@cmungall the access note in bootstrap_client.sh is a shell comment, so it never reaches the terminal. Most people will run this through `uv run beril setup`, which calls the script in a subprocess and on failure prints "Fix the issue above and retry". What they see above is pip's `Repository not found` and nothing about access or #407. A plain `echo` before the pip install would surface it:

    echo "NOTE: spark_connect_remote and berdl_remote are private repos in the KBaseDataLakehouse org." >&2
    echo "      If the install below fails with 'Repository not found', request read access via #407." >&2

Also, `.gitignore` lines 16-17 still say the packages are pinned from github.com/BERDataLakehouse and that the org returns 404. Worth updating in the same pass so the explanation there matches the new one (private, not gone).

PR #435, opened 2026-09-17
Make /api/context/ls addressable, scoped to the caller

briehl:
## Summary

`GET /api/context/ls` took **no arguments at all** — hardcoded to the shared `resources/projects/` root — so it could not serve `ls <path>`, recursive listings, or path-only output. This makes it addressable.

Unlike the rest of the parity work (A1, #427), this is a **behavior change** rather than an addition. Only tests consumed the route, so nothing in the app breaks; three tests that asserted the hardcoded contract were rewritten.

Part A2 of the context-API parity work. Stacked on #427.

## The design decision, and why it isn't a `uri` parameter

Checking consumers surfaced a security dimension worth stating plainly:

- **Ingest writes to per-user namespaces** — `viking://resources/users/<orcid>/<project_slug>/…`
- **`/ls` read the shared `resources/projects/` root**, and took no path

So the moment this route accepts a caller-supplied path, an arbitrary path means **any user can list any other user's ingested files**.

The route therefore takes `project` + `path` and resolves the namespace from the **authenticated ORCiD**. A `uri` parameter would have to be validated against the caller's own prefix, and one missed check would expose someone else's data. This way the API cannot express it at all.

`listing_uri()` is where that guarantee lives: the ORCiD prefix is structural, and a traversal that would climb out of it raises rather than resolving.

## API

```
GET /api/context/ls                                    -> the caller's projects
GET /api/context/ls?project=alpha                      -> that project
GET /api/context/ls?project=alpha&path=memories        -> a subpath
GET /api/context/ls?project=alpha&recursive=true&simple=true
GET /api/context/ls?node_limit=50
```

- `path` without `project` is **422** — it would otherwise resolve against the namespace root and list across projects.
- An **un-ingested project lists empty, not 404**. That is a legitimate state, and distinguishing it would report on a namespace the caller may not have written yet.
- `node_limit` is bounded (1–10,000) like `find`'s, so a request cannot ask the backend to walk its whole graph.
- Backend failures surface as **502** via the `QUERY_FAILURES` tuple added in #427.
- Both `list_files` layers now close the client when the call raises — neither did before.

## Testing

26 new cases across both layers (972 webapp + 769 root tests pass).

Security coverage specifically:
- `../<other-orcid>`, `../../users/other`, `..`, and `a/../../escape` are all 422 before the backend is touched
- a caller-supplied `orcid` query parameter is ignored — the session's ORCiD is used
- two users' identically named projects never resolve to the same URI

**I verified these tests can actually fail.** Replacing the traversal-checked join in `listing_uri` with a naive f-string breaks 6 tests across both layers, including the `../<other-orcid>` case — so they exercise the real check, not a tautology.

## Follow-up this opens

The shared `resources/projects/` corpus that `ingest_context.py` populates is now **unreachable through `/ls`**. Nothing in `.claude` reads it via this route today, but `knowledge_query.py`'s `ls viking://resources/projects/` examples do — so Phase B needs either a `scope=shared` option or those examples rewritten. Flagging rather than guessing; it's a decision for A3-onward.


dileep-kishore:
Commenting here and on #437 since the two go together. The second commit in #437 changes the access model both descriptions argue for: `owner` and `all_owners` now permit reads across every namespace. The old "orcid parameter is ignored" test still passes but no longer demonstrates caller-only reads. That's a defensible model for submitted projects, but please update both descriptions or squash the rework into this PR, and decide what `UserProject.is_public` means now, since ingest creates rows with it false and reads ignore it. One gap: there's no way to ask for "project alpha, whoever owns it", since `all_owners` drops the project name.

